PR #2, opened 2017-06-22
SLVSCODE-6, SLVSCODE-13 Update analyzers and improve package script

henryju:


scripts/prepare.js
```diff
@@ -0,0 +1,42 @@
...
+if (!fs.existsSync('analyzers')){
+    fs.mkdirSync('analyzers');
+}
+
+downloadIfNeeded('https://repox.sonarsource.com/sonarsource-public-builds/org/sonarsource/sonarlint/core/sonarlint-language-server/2.15.0.780/sonarlint-language-server-2.15.0.780.jar', 'server/sonarlint-ls.jar');
+downloadIfNeeded('http://repo1.maven.org/maven2/org/sonarsource/javascript/sonar-javascript-plugin/3.1.0.5111/sonar-javascript-plugin-3.1.0.5111.jar', 'analyzers/sonarjs.jar');
+downloadIfNeeded('http://repo1.maven.org/maven2/org/sonarsource/php/sonar-php-plugin/2.10.0.2087/sonar-php-plugin-2.10.0.2087.jar', 'analyzers/sonarphp.jar');
+downloadIfNeeded('http://repo1.maven.org/maven2/org/sonarsource/python/sonar-python-plugin/1.8.0.1496/sonar-python-plugin-1.8.0.1496.jar', 'analyzers/sonarpython.jar');
+
+function downloadIfNeeded(url, dest) {
```
janos-ss: This will download the sha1 every time. But the sha1 is not expected to change, normally. So I think that's a waste. 

I think it would be better to simply check if the file exists. Download if it doesn't. And then of course, when downloading a jar, it's good to download the sha1 as well and verify the checksum.

henryju: The point is that the file may already exists, even if you change the URL. So if you forgot to clean, you'll end up using an outdated version. That's why I'm always checking checksum.

src/extension.ts
```diff
@@ -39,6 +39,9 @@ function runJavaServer(context: VSCode.ExtensionContext) : Thenable<StreamInfo>
 				let vmargs = getSonarLintConfiguration().get('ls.vmargs','');
 				parseVMargs(params, vmargs);
 				params.push('-jar', serverJar, '' + languageServerPort);
+				params.push(toUrl(Path.resolve(context.extensionPath, "analyzers", "sonarjs.jar")));
+				params.push(toUrl(Path.resolve(context.extensionPath, "analyzers", "sonarphp.jar")));
+				params.push(toUrl(Path.resolve(context.extensionPath, "analyzers", "sonarpython.jar")));
```
janos-ss: It's not great that this file and `scripts/prepare.js` both use the hardcoded names for the jars. When updating at one place, you have to remember to update the other too, and make sure the names match precisely. Isn't there a way to have these names in a shared list, which can be updated in a single place? 

If that's not possible, then to mitigate the risk, at least it would be good to have these names in an easy-to-find place, for example in a lists defined near the top of the files, along with a comment referencing the other file that needs to be updated together.

scripts/prepare.js
janos-ss: I see. Then it's fine, I don't have a better idea.

src/extension.ts
henryju: Let say this is intentional. Another option would have been to list files in the "analyzers" folder, but I'm a bit afraid of adventurous users trying to drop random analyzers in this folder.
And since the file names doesn't contain the version, we should not have to update them.

PR #4, opened 2017-06-26
Feature/open desc

henryju:


package.json
```diff
@@ -14,7 +14,7 @@
     "url": "https://jira.sonarsource.com/browse/SLVSCODE",
     "email": "sonarlint@googlegroups.com"
   },
-  "license": "LGPL-3.0",
+  "license": "SEE LICENSE IN LICENSE.txt",
```
janos-ss: I think "See LICENSE.txt" would be enough

henryju: @janos-ss See https://code.visualstudio.com/docs/extensionAPI/extension-manifest

janos-ss: Interesting, good to know!

henryju: Note that this page is specific to vscode, not sure it is standard for all npm modules.

PR #6, opened 2018-01-23
SLVSCODE-21 Enable TypeScript analysis

henryju:


package.json
```diff
@@ -87,17 +93,17 @@
     "server/sonarlint-ls.jar"
   ],
   "scripts": {
-    "vscode:prepublish": "tsc -p ./ && node scripts/prepare.js",
-    "compile": "tsc -watch -p ./",
+    "vscode:prepublish": "./node_modules/.bin/tsc -p ./ && node scripts/prepare.js",
+    "compile": "./node_modules/.bin/tsc -watch -p ./",
```
vilchik-elena: @henryju do we need this `watch`?

scripts/prepare.js
```diff
@@ -27,6 +27,10 @@ downloadIfNeeded(
   "http://repo1.maven.org/maven2/org/sonarsource/python/sonar-python-plugin/1.8.0.1496/sonar-python-plugin-1.8.0.1496.jar",
   "analyzers/sonarpython.jar"
 );
+downloadIfNeeded(
+  "https://repox.sonarsource.com/sonarsource/org/sonarsource/typescript/sonar-typescript-plugin/1.5.0.2021/sonar-typescript-plugin-1.5.0.2021.jar",
```
vilchik-elena: @henryju we will update these build numbers (ts and sonarlint-ls) before release, right?

src/extension.ts
```diff
@@ -126,18 +131,61 @@ function toUrl(filePath) {
...
+
+  return Path.join(VSCode.workspace.rootPath, tsdkPathSetting);
+}
+
 export function activate(context: VSCode.ExtensionContext) {
   let serverOptions = () => runJavaServer(context);
 
+  const tsExt = VSCode.extensions.getExtension("vscode.typescript");
+  if (!tsExt) {
+    console.warn(
```
vilchik-elena: @henryju should we return after this `warn`?

src/extension.ts
```diff
@@ -126,18 +131,61 @@ function toUrl(filePath) {
...
+      "Unable to locate TypeScript extension. No TypeScript support in SonarLint"
+    );
+  }
+  const tsdkPathSetting = VSCode.workspace
+    .getConfiguration("typescript")
+    .get("tsdk", undefined);
```
vilchik-elena: @henryju omit second parameter, it will do the same job

package.json
henryju: No idea. It certainly come from the original template we used as a quick starter. Do you suggest we should remove it?

vilchik-elena: @henryju in fact it does nothing:) it should be `-w` or `--watch`. So you can remove it.

PR #9, opened 2018-01-25
SLVSCODE-17 Add support for Vue.js

henryju:


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "sonarlint-vscode",
   "displayName": "SonarLint",
   "description":
-    "On-the-fly bug and quality issues detection for JavaScript, Python and PHP",
+    "On-the-fly bug and quality issues detection for JavaScript, TypeScript, Python and PHP",
```
vilchik-elena: @henryju should it be in different commit?

henryju: @vilchik-elena Indeed

PR #11, opened 2018-01-26
Add Typescript support and rules section

juboucher:


DanielRuf:
How can we enable rules?

henryju:
For VSCode there is currently no way to configure rules without using connected mode with a SonarQube server. This is something we should improve this year.
https://jira.sonarsource.com/browse/MMF-1277

DanielRuf:
Would be great to have this feature in the future.

DanielRuf:
Because the rule for console.log is not used in SonarLint for VSCode as it seems (I use SonarScanner 3.3.0).

PR #12, opened 2018-02-12
Please enable GitHub issues

paulvi:
Google Groups and all Google services are not available in China (without VPN)

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ SonarLint should automatically find it but you can also explicitely set the path
 
 ## Contributions and license
 
-SonarLint for Visual Studio Code is open source under the LGPL v3 license. Feel free to submit Pull Requests.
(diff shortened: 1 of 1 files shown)


henryju:
We use JIRA as issue tracker, so we don't want to enable GitHub issues. We are not 100% satisfied with Google Group, so we are experimenting a new way to exchange with the community. Stay tuned.

paulvi:
OK, then add URL to JIRA in README

henryju:
The README is what is displayed on the marketplace page, so we try to keep it more functional than technical. The bug tracker is referenced in the package.json (https://github.com/SonarSource/sonarlint-vscode/blob/17c58bc3170f939bf74372fa74cdad961bdfb726/package.json#L15) and that make it appear on the right panel.

PR #29, opened 2018-07-31
Add possibility to use relative path in javaHome setting

Franziskus1988:
Hi, in our company we have a java jre in our repo and share the workspace settings of vscode. Currently it is not possible to set a relative path from vscode workspace folder to the java jre, because the extension always expects the javaHome path to be absolute. Or at least the extension says it can't find jre folder. I'd like to fix it with this changes.

### DIFF
--- a/src/requirements.ts
+++ b/src/requirements.ts
@@ -39,6 +39,9 @@ function checkJavaRuntime(): Promise<any> {
     let source: string;
     let javaHome: string = readJavaConfig();
     if (javaHome) {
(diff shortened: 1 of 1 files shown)


henryju:
I'm closing this since having a JRE in a SCM repository is not common (to say the least).

Franziskus1988:
@henryju Häää. This doesn't make sense at all. The problem is use case independant. It's just not possible to use a relative path for this setting. Why shouldn't add this 3 lines, which in my opinion shouldn't hurt?

henryju:
Another reason of why I'm not confident to merge this: in a multi-folder situation, why do you choose the first folder to resolve the relative path? What if you have a different JRE in each folder?

PR #32, opened 2019-02-18
Hardening + SonarHTML

henryju:


package.json
```diff
@@ -43,92 +46,78 @@
...
         },
         "sonarlint.ls.javaHome": {
-          "type": [
-            "string",
-            "null"
-          ],
-          "default": null,
-          "description": "Specifies the folder path to the JRE (8 or more recent) used to launch the SonarLint Language Server.\nOn Windows, backslashes must be escaped, i.e.\n\"sonarlint.ls.javaHome\":\"C:\\\\Program Files\\\\Java\\\\jdk1.8.0_161\"",
+          "type": "string",
+          "markdownDescription": "Path to a JRE (8 or more recent) used to launch the SonarLint Language Server. \n\n On Windows, backslashes must be escaped, e.g. `C:\\\\Program Files\\\\Java\\\\jdk1.8.0_161`",
```
janos-ss: I'm wondering if using `/` would work. I would much prefer writing like that. I can test in Windows tomorrow, if you like.

package.json
```diff
@@ -43,92 +46,78 @@
...
                 "default": ""
               },
               "serverUrl": {
                 "type": "string",
+                "description": "URL of the server. Use https://sonarcloud.io for SonarCloud.",
                 "default": ""
               },
               "token": {
                 "type": "string",
+                "description": "Token generated from My Account>Security in SonarQube/SonarCloud",
```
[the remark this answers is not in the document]
henryju: I have no opinion, let see what @christophelevis think of it.

package.json
henryju: Please test and tell me. But I found other extensions doing the same, so I have doubt.

janos-ss: Tested, using `/` works on Windows. Even in earlier versions of VSCode (I found my settings on Windows were always using `/`).

Anyway, let's keep it as it is. Perhaps more natural with `\` for Windows users, and as you said other plugins are doing the same.

PR #28, opened 2018-07-06
Adds Java support

morganfogg:
Adds support for the Java programming language. Tested on several projects, found no issues so far. Screenshot below.

screenshot_20180706_103913

PR #33, opened 2019-02-26
SLVSCODE-35 Add basic java support

badsyntax:
See https://jira.sonarsource.com/browse/SLVSCODE-35

In an effort to get some movement on that feature request, i've had a go at adding java support.

There's not many changes but it seems to be work. I have not really done much testing and i'm not sure how TBH. 

I used the latest version found via https://mvnrepository.com/artifact/org.sonarsource.java/sonar-java-plugin/5.11.0.17289

Here is a screenshot showing it working:

screenshot 2019-02-26 at 16 30 40


badsyntax:
Ok i've just seen https://github.com/SonarSource/sonarlint-vscode/pull/28 which is almost exactly the same.

PR #36, opened 2019-04-25
Update JRE requirement (at 1.5 we support JRE 10)

janos-ss:


README.md
```diff
@@ -25,7 +25,7 @@ You will benefit from the following code analyzers: [SonarJS](https://redirect.s
 
 ## Requirements
 
-The only thing you need is a Java Runtime (JRE) 8 installed on your computer.
+The only thing you need is a Java Runtime (JRE) 8, 9 or 10 installed on your computer.
```
henryju: Don't we support 11? If not, that would deserve a ticket IMO.

janos-ss:
On a quick look, indeed it seems to work fine with Java 11. I think we can leave the README alone. I close this PR now.

PR #39, opened 2019-05-19
Add sonarlint.debug config

Idered:
Problem:
I'm working on custom vs code extension and sonarlint is spamming my Debug Console.

Solution:
Added `solarlint.debug` config. It's set to false by default.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -33,6 +33,8 @@ const connectedModeServersSectionName = 'connectedMode.servers';
 const connectedModeProjectSectionName = 'connectedMode.project';
 
 function runJavaServer(context: VSCode.ExtensionContext): Thenable<StreamInfo> {
+  const enableDebug = getSonarLintConfiguration().get('debug')
+
   return requirements
     .resolveRequirements()
     .catch(error => {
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -33,6 +33,8 @@ const connectedModeServersSectionName = 'connectedMode.servers';
 const connectedModeProjectSectionName = 'connectedMode.project';
 
 function runJavaServer(context: VSCode.ExtensionContext): Thenable<StreamInfo> {
+  const enableDebug = getSonarLintConfiguration().get('debug')
```
henryju: This will only be evaluated at startup right? Would be better to support a dynamic change of this property, because sometimes restarting the IDE make it difficult to reproduce an issue.

henryju:
I think it has been fixed by those commits:
https://github.com/SonarSource/sonarlint-vscode/commit/d2731d7df9b8ebef095077e3d987e442f7001bd1#diff-45327f86d4438556066de133327f4ca2R80
https://github.com/SonarSource/sonarlint-core/commit/11a68962b9c4624a908adc29d1b5d6212a7738a0

PR #40, opened 2019-06-21
SLVSCODE-68 Update language server to fix storage issue

henryju:


scripts/prepare.js
```diff
@@ -3,7 +3,7 @@ const fs = require('fs');
 const crypto = require('crypto');
 const request = require('request');
 
-const languageServerVersion = '4.2.0.2266';
+const languageServerVersion = '4.3.0.2409';
```
jblievremont: Should probably be updated to `4.3.0.2416` to use latest PR artifact.

PR #46, opened 2019-07-09
Update changelog and readme for 1.9 release

jblievremont:


README.md
```diff
@@ -41,6 +41,8 @@ To configure the connection, have a look at SonarLint in default user settings.
 
 If you change something on the server such as the quality profile, you can trigger an update of the local cache using the "Update SonarLint binding to SonarQube/SonarCloud" command on the command palette (search for "sonarlint").
 
+When several sub-folders of the same server project are open in a single workspace, the project binding can be set at the workspace level.
```
christophelevis: I guess we never really recommend to configure the binding at _User_ level.
What about saying  something like this?
'Configuring the project binding at workspace level mutes Won’t Fix or False Positive issues in any of the sub-folders you add to the workspace.'

I would then suggest to write it just after:
'To configure the connection, have a look at SonarLint in default user settings.'

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 1.9.0
+
+* Enable support of PL/SQL in connected mode
+* Enable connected mode for sub-folders and multi-folder workspaces (single project)
```
christophelevis: 'Support for connected mode on all sub-folders of a same project in a workspace'?

README.md
```diff
@@ -25,12 +25,12 @@ You will benefit from the following code analyzers: [SonarJS](https://redirect.s
 
 ## Requirements
 
-The only thing you need is a Java Runtime (JRE) 8 installed on your computer.
+The only thing you need is a Java Runtime (JRE) 8 to 11 installed on your computer.
```
christophelevis: 'or 11'
I don't think we should officially support intermediary versions.

PR #47, opened 2019-07-11
Fix language server freeze after a few analyses

jblievremont:
a2745cbbab0b46a25088d66d5038f2829aef1d69 introduced a regression: since the standard streams of the forked process were not consumed, the buffer got full on language server side and hung analyses.

This new version unconditionally attaches stream consumers, then only logs conditionally.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -35,7 +35,7 @@ let languageClient: LanguageClient;
 
 function logToSonarLintOutput(message) {
   if (sonarlintOutput) {
-    sonarlintOutput.append(message);
+    sonarlintOutput.appendLine(message);
   }
 }
 
@@ -54,7 +54,7 @@ function runJavaServer(context: VSCode.ExtensionContext): Thenable<StreamInfo> {
     .then(requirements => {
       return new Promise<StreamInfo>((resolve, reject) => {
         const server = Net.createServer(socket => {
-          console.log(`Child process connected on port ${server.address().port}`);
+          logToSonarLintOutput(`Child process connected on port ${server.address().port}`);
           resolve({
             reader: socket,
             writer: socket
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -63,14 +70,12 @@ function runJavaServer(context: VSCode.ExtensionContext): Thenable<StreamInfo> {
           console.log(`Executing ${command} ${args.join(' ')}`);
```
henryju: Replace all console.log IMO (and use appendLine)

src/extension.ts
```diff
@@ -464,8 +470,16 @@ function getSonarLintConfiguration(): VSCode.WorkspaceConfiguration {
...
-    return undefined;
+  if (languageClient) {
+    return languageClient.stop()
+      .then(disposeOutput);
   }
-  return languageClient.stop();
+  return Promise.resolve(disposeOutput());
+}
+
+function disposeOutput() {
```
henryju: Why not using `context.subscriptions.push(sonarlintOutput);` in the activate instead?

jblievremont: Good catch, I had not seen this in the docs.

src/extension.ts
```diff
@@ -60,17 +67,15 @@ function runJavaServer(context: VSCode.ExtensionContext): Thenable<StreamInfo> {
...
-          if (VERBOSE_LOGS) {
-            process.stdout.on('data', function(data) {
-              console.log(data.toString());
-            });
-            process.stderr.on('data', function(data) {
-              console.error(data.toString());
-            });
-          }
+          process.stdout.on('data', function(data) {
+            logToSonarLintOutput(data.toString());
```
henryju: Here I'm not sure we should use appendLine, since data will probably contains carriage returns (to be verified)

jblievremont: Yup, confirmed with experiment. I also believe that showing the raw `stdout`/`stderr` content (instead of adding newline characters) will make troubleshooting easier.

PR #48, opened 2019-08-12
MMF-1277 Allow users to disable rules

jblievremont:


gulpfile.js
```diff
@@ -50,32 +51,35 @@ gulp.task('compute-hashes', ['package'], function() {
 });
 
 gulp.task('deploy-vsix', ['package', 'compute-hashes'], function() {
-  if (process.env.TRAVIS_BRANCH != 'master') {
+  if (process.env.TRAVIS_BRANCH !== 'master') {
```
siegfried-ehret-sonarsource: Are you sure you don't need one or two other `=` signs ? We never know, it's JS... :trollface:

jblievremont: :man_facepalming:

src/extension.ts
```diff
@@ -255,6 +254,15 @@ export function activate(context: VSCode.ExtensionContext) {
     }
   );
 
+  VSCode.commands.registerCommand('SonarLint.DeactivateRule', (ruleKey: string) => {
+    const configuration = getSonarLintConfiguration();
+    const rules = configuration.has('rules') ? configuration.get('rules') : {};
+    rules[ruleKey] = {
+      level: 'off'
+    };
+    return configuration.update('rules', rules, VSCode.ConfigurationTarget.Global);
```
siegfried-ehret-sonarsource: What is the purpose of `VSCode.ConfigurationTarget.Global` ?
I see in the file that we get the SonarLint config from the workspace `VSCode.workspace.getConfiguration('sonarlint')`, there is no equivalent to update the config (in the workspace instead of Global) ? Or maybe workspace and global are the same thing ?

jblievremont: This asymmetry comes from the API itself: the configuration is read from the `workspace` object, which merges the user, workspace and folder configurations (when appropriate). On write, it's up to the extension to choose where to store the settings - and in the scope of this sprint, we want to store it in the global user settings.

package.json
```diff
@@ -67,6 +67,25 @@
           "default": false,
           "markdownDescription": "Disable sending anonymous usage statistics"
         },
+        "sonarlint.rules": {
+          "type": "object",
+          "scope": "application",
+          "default": {},
+          "markdownDescription": "Configure rules. In connected mode, this configuration is overridden by the server's quality profile.\n\nExample:\n\n    \"sonarlint.rules\": {\n        \"javascript:UnusedVariable\": {\n            \"level\": \"off\"\n        }\n    }\n",
```
henryju: I would say 
> In connected mode, this configuration is overridden by the **project**'s quality profile`

PR #50, opened 2019-08-19
SLCORE-244 Trigger diagnostics refresh on rules configuration change

jblievremont:
Client side of SLCORE-244, using dedicated command to trigger a refresh of diagnostics after an update to rule configuration.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -274,7 +274,7 @@ export function activate(context: VSCode.ExtensionContext) {
   });
 
   VSCode.workspace.onDidChangeConfiguration(async event => {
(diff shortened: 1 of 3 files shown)


src/extension.ts
```diff
@@ -271,6 +273,23 @@ export function activate(context: VSCode.ExtensionContext) {
       });
   });
 
+  VSCode.workspace.onDidChangeConfiguration(async event => {
+    if(event.affectsConfiguration("sonarlint.rules")) {
```
henryju: formating issue (missing space between if and (

PR #52, opened 2019-09-11
SLVSCODE-79 Describe "on" value for rule sonarlint.rules settings

jblievremont:


jblievremont:
Will be merged as part of #55

PR #55, opened 2019-09-16
SLVSCODE-80 Show list of rules in dedicated view

jblievremont:


src/util.ts
```diff
@@ -4,7 +4,7 @@
  * sonarlint@sonarsource.com
  * Licensed under the LGPLv3 License. See LICENSE.txt in the project root for license information.
  * ------------------------------------------------------------------------------------------ */
-'use strict';
+"use strict";
```
henryju: Nope, keep single quotes :)

PR #56, opened 2019-09-17
Improve Java home detection

henryju:
Largely inspired from https://github.com/redhat-developer/vscode-java/commit/6bccceab4eda659d03c149daf17a686f85d70158#diff-baff02b04010a735985f63100efb9fc2

- display detected path
- check if java home ends with `bin`
- opens settings if sonarlint.ls.javaHome preference is invalid

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -209,7 +209,8 @@ export function activate(context: VSCode.ExtensionContext) {
   oldConfig = getSonarLintConfiguration();
 
   // Create the language client and start the client.
-  languageClient = new LanguageClient('sonarlint-vscode', 'SonarLint Language Server', serverOptions, clientOptions);
+  // id parameter is used to load 'sonarlint.trace.server' configuration
+  languageClient = new LanguageClient('sonarlint', 'SonarLint Language Server', serverOptions, clientOptions);
 
   VSCode.commands.registerCommand(
     'SonarLint.OpenRuleDesc',
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -63,8 +63,8 @@ function runJavaServer(context: VSCode.ExtensionContext): Thenable<StreamInfo> {
     .catch(error => {
       //show error
       VSCode.window.showErrorMessage(error.message, error.label).then(selection => {
-        if (error.label && error.label === selection && error.openUrl) {
-          VSCode.commands.executeCommand('vscode.open', error.openUrl);
+        if (error.label && error.label === selection && error.command) {
+          VSCode.commands.executeCommand(error.command, error.commandParam);
         }
       });
       // rethrow to disrupt the chain.
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -191,6 +191,7 @@ export function activate(context: VSCode.ExtensionContext) {
       return {
         testFilePattern: configuration && configuration.get('testFilePattern'),
         analyzerProperties: configuration && configuration.get('analyzerProperties'),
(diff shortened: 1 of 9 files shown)


src/requirements.ts
```diff
@@ -13,7 +13,8 @@ import * as cp from 'child_process';
 import * as path from 'path';
 import * as pathExists from 'path-exists';
 import * as expandHomeDir from 'expand-home-dir';
-import * as findJavaHome from 'find-java-home';
+import findJavaHome = require('find-java-home');
```
jblievremont:
```suggestion
import * as findJavaHome from 'find-java-home';
```
Why use require instead of import here?

src/requirements.ts
```diff
@@ -13,7 +13,8 @@ import * as cp from 'child_process';
...
+import { Commands } from './commands';
```
jblievremont: I see what you meant by conflicts :innocent:

src/requirements.ts
```diff
@@ -26,8 +27,8 @@ export interface RequirementsData {
 interface ErrorData {
   message: string;
   label: string;
-  openUrl: Uri;
-  replaceClose: boolean;
+  command: string;
```
jblievremont: TypeScript tells me this `ErrorData` interface is not used anymore, we might as well remove it.

src/requirements.ts
henryju: No idea, copied the same change they made in vscode-java. I though it might be important, but if you are confident it is harmless I'm ok to revert.

src/extension.ts
```diff
@@ -63,8 +63,8 @@ function runJavaServer(context: VSCode.ExtensionContext): Thenable<StreamInfo> {
     .catch(error => {
       //show error
       VSCode.window.showErrorMessage(error.message, error.label).then(selection => {
-        if (error.label && error.label === selection && error.openUrl) {
-          VSCode.commands.executeCommand("vscode.open", error.openUrl);
+        if (error.label && error.label === selection && error.command) {
+          VSCode.commands.executeCommand(error.command, error.commandParam);
```
henryju: Here is the place where `ErrorData` should be useful, but for some reason the type is not resolved.

henryju:
@jblievremont what about the change of quotes to double quotes? Have you sorted out why the prettier configuration is not fixed?

src/requirements.ts
henryju: I applied you fix

src/requirements.ts
```diff
@@ -36,28 +37,33 @@ export async function resolveRequirements(): Promise<RequirementsData> {
...
     }
     if (javaHome) {
       javaHome = expandHomeDir(javaHome);
       if (!pathExists.sync(javaHome)) {
-        openJREDownload(reject, source + ' points to a missing folder');
-      }
-      if (!pathExists.sync(path.resolve(javaHome, 'bin', JAVA_FILENAME))) {
-        openJREDownload(reject, source + ' does not point to a JRE.');
+        invalidJavaHome(reject, `The ${source} points to a missing or inaccessible folder (${javaHome})`);
+      } else if (!pathExists.sync(path.resolve(javaHome, 'bin', JAVA_FILENAME))) {
```
Ekhozie: With this check we cannot use Windows symbolic link path, for instance: `C:\ProgramData\Oracle\Java\javapath` or `C:\Program Files (x86)\Common Files\Oracle\Java\javapath` which force us to update vscode configuration each team we update the JRE

PR #65, opened 2019-11-20
Supports binding per workspace folder

henryju:


gulpfile.js
```diff
@@ -77,7 +78,7 @@ gulp.task('deploy-vsix', ['package', 'compute-hashes'], function() {
         password: ARTIFACTORY_DEPLOY_PASSWORD,
         properties: {
           'vcs.revision': BUILD_SOURCEVERSION,
-          'vcs.branch': BUILD_SOURCEBRANCH,
+          'vcs.branch': SYSTEM_PULLREQUEST_TARGETBRANCH || BUILD_SOURCEBRANCH,
```
jblievremont:
```suggestion
          'vcs.branch': SYSTEM_PULLREQUEST_SOURCEBRANCH || BUILD_SOURCEBRANCH,
```

src/extension.ts
```diff
@@ -308,16 +298,6 @@ export function activate(context: VSCode.ExtensionContext) {
     })
   );
 
-  context.subscriptions.push(
-    VSCode.commands.registerCommand(Commands.UPDATE_SERVERS_AND_BINDING_STORAGE, () => {
-      updateServerStorage()
-        .then(updateProjectBinding)
-        .then(() => {
-          VSCode.window.showInformationMessage('SonarLint server storage updated');
```
jblievremont: If I understand correctly, users will now have to look at their LS logs to get feedback. I think we should try to find a way to provide more high level feedback about synchronization progress.

gulpfile.js
henryju: I disagree. If you look at the [parent pom](https://github.com/SonarSource/parent-oss/commit/ea6a1dc54f41ac922f458fcf802bb752b0af4c8e#diff-600376dffeb79835ede4a0b285078036R676), the variable was `TRAVIS_BRANCH` which for a PR is the target branch.

src/extension.ts
henryju: We can show a message from the LS to the UI at the end of the process. I will have a look.

PR #69, opened 2020-01-15
SLVSCODE-90 Add QA

henryju:


its/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "sonarlint-its",
+  "displayName": "sonarlint-its",
+  "description": "SonarLint ITs",
+  "version": "0.0.1",
+  "engines": {
+    "vscode": "^1.41.0"
```
jblievremont: Should this be `1.36.0` for consistency with the extension under test?

henryju: Yep, but this is surprising the IT axis 1.36.0 was passing...

PR #73, opened 2020-01-29
SLVSCODE-107 Offer to automatically install a JRE when none is detected

jblievremont:


package.json
```diff
@@ -285,15 +285,18 @@
   "dependencies": {
     "expand-home-dir": "0.0.3",
     "find-java-home": "1.0.1",
+    "follow-redirects": "1.10.0",
```
henryju: License MIT :+1:

package.json
```diff
@@ -285,15 +285,18 @@
...
+    "inly": "4.0.4",
```
henryju: License MIT :+1:

src/requirements.ts
```diff
@@ -66,33 +59,54 @@ function checkJavaRuntime(): Promise<string> {
...
         invalidJavaHome(reject, msg);
       }
-      return resolve(javaHome);
+      resolve(javaHome);
     }
-    //No settings, let"s try to detect as last resort.
+    // No settings, check if we have a managed one
+    javaHome = context.globalState.get(JAVA_MANAGED_HOME_KEY, null);
+    if (javaHome) {
+      resolve(javaHome);
```
siegfried-ehret-sonarsource: Maybe (not sure) you need to add a return, or your program here will continue and you may call `resolve` or `reject` after the promise has been `resolve`d here.

src/requirements.ts
```diff
@@ -66,33 +59,54 @@ function checkJavaRuntime(): Promise<string> {
...
 function readJavaConfig(): string {
-  const config = workspace.getConfiguration();
-  return config.get<string>(JAVA_HOME_CONFIG, null);
+  return vscode.workspace.getConfiguration().get<string>(JAVA_HOME_CONFIG, null);
 }
 
 function checkJavaVersion(javaHome: string): Promise<number> {
   return new Promise((resolve, reject) => {
-    cp.execFile(javaHome + '/bin/java', ['-version'], {}, (error, stdout, stderr) => {
+    cp.execFile(`${javaHome}/bin/java`, ['-version'], {}, (error, stdout, stderr) => {
```
siegfried-ehret-sonarsource: You used `path.join` somewhere else, could be nice to have it here to (I don't know what happens behind `execFile`).

src/requirements.ts
```diff
@@ -145,3 +168,37 @@ function invalidJavaHome(reject, cause: string) {
     });
   }
 }
+
+export async function installManagedJre() {
+  return vscode.window.withProgress({ location: vscode.ProgressLocation.Notification, title: 'Downloading JRE' },
+    async (progress, cancelToken) => {
+      const platformInfo = await PlatformInformation.GetPlatformInformation();
```
siegfried-ehret-sonarsource: When an error happens in an `await`ed thing, an exception is thrown. You can add a try/catch.

src/jre.ts
```diff
@@ -0,0 +1,114 @@
...
+
+const https = followRedirects.https;
+
+const ADOPT_OPEN_JDK_API_ROOT = 'https://api.adoptopenjdk.net/v2';
+
+type RequestType = 'info' | 'binary' | 'latestAssets';
+type ReleaseType = 'releases' | 'nightly';
+export type Version = 8 | 9 | 10 | 11 | 12 | 13;
+type Implementation = 'hotspot' | 'openj9';
+export type Os = 'windows' | 'linux' | 'mac' | 'solaris' | 'aix';
```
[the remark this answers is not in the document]
jblievremont: This is the full range of values supported by the AdoptOpenJDK API, I put those here mainly for documentation purposes.

src/requirements.ts
```diff
@@ -145,3 +170,39 @@ function invalidJavaHome(reject, cause: string) {
...
+  return vscode.window.withProgress({ location: vscode.ProgressLocation.Notification, title: 'Downloading JRE' },
+    async (progress, cancelToken) => {
+      const options = await PlatformInformation.GetPlatformInformation();
+      console.log(`Detected options ${JSON.stringify(options)}`);
+      return jre.download({
+        os: options.os as jre.Os,
+        architecture: options.arch as jre.Architecture,
+        version: 11
+      }, path.join(util.extensionPath, '..', 'sonarsource.sonarlint_managed-jre'))
+        .then((downloadResponse: any) => {
```
wouter-admiraal-sonarsource: How about typing the response to an interface instead of using `any`?

src/jre.ts
```diff
@@ -0,0 +1,111 @@
...
+    `os=${actualOptions.os}`,
+    `arch=${actualOptions.architecture}`,
+    `type=${actualOptions.binary}`,
+    `heap_size=${actualOptions.heapSize}`,
+    `release=latest`
+  ].join('&');
+  return `${requestUrl}?${requestParams}`;
+}
+
+export function download(options: Options, destinationDir: string) {
```
wouter-admiraal-sonarsource: How about typing the result of this as `Promise<{ destinationDir: string; jreZipPath : string; }>`?

src/linuxDistribution.ts
```diff
@@ -0,0 +1,67 @@
...
+        // as per the os-release documentation.
+        linuxDistro = LinuxDistribution.getDistroInformationFromFile('/etc/os-release')
+            .catch(() => {
+                return LinuxDistribution.getDistroInformationFromFile('/usr/lib/os-release');
+            }).catch(() => {
+                // couldn't get distro information
+                return Promise.resolve(new LinuxDistribution('unknown', 'unknown'));
+            });
+
+        return linuxDistro;
```
wouter-admiraal-sonarsource: Any reason for assigning this to a local `let` instead of returning immediately?

src/linuxDistribution.ts
```diff
@@ -0,0 +1,67 @@
...
+                    reject(error);
+                    return;
+                }
+                resolve(LinuxDistribution.getDistroInformation(data));
+            });
+        });
+    }
+
+    // Only public for tests.
+    public static getDistroInformation(data: string): LinuxDistribution {
```
wouter-admiraal-sonarsource: I understand why this is here, but still... I find it a bit weird to have a _public_ method, with the **same name** (except the case of the first char) as another one, just for tests... This might lead to confusion down the road.

If this really needs to remain public, but you want to make it extra clear this is _not_ to be used, how about prefixing it with an `_` (underscore), the old-school way?

src/platform.ts
```diff
@@ -0,0 +1,81 @@
...
+        'darwin': 'mac'
+    }[platform];
+}
+
+function archToJreArch(arch: string) {
+    return {
+        'x86_64': 'x64',
+        'x86': 'x32'
+    }[arch];
+}
```
wouter-admiraal-sonarsource: Nitpicking: slight risk of getting an undefined here. How about using an `enum`, and/or restricting what can be passed as an argument (using the enum keys, or typing them out explicitly)?

src/platform.ts
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        return architecturePromise.then(arch => {
+            return new PlatformInformation(platformToJreOs(platform), archToJreArch(arch));
+        });
+    }
+
+    public static GetUnknownArchitecture(): string { return 'Unknown'; }
+
+    private static async GetWindowsArchitecture(): Promise<string> {
```
wouter-admiraal-sonarsource: Nitpicking: You define this as an `async` function, but there isn't any `await`. Is that intended?

src/platform.ts
```diff
@@ -0,0 +1,81 @@
...
+                        }
+                    }
+                }
+                return PlatformInformation.GetUnknownArchitecture();
+            }).catch((error) => {
+                return PlatformInformation.GetUnknownArchitecture();
+            });
+    }
+
+    private static async GetUnixArchitecture(): Promise<string> {
```
wouter-admiraal-sonarsource: Same as above.

wouter-admiraal-sonarsource:
Ah sorry, I see I commented on a specific commit rather than the final version 🤦‍♂.

src/requirements.ts
```diff
@@ -29,30 +34,18 @@ interface ErrorData {
   message: string;
   label: string;
   command: string;
-  commandParam: any;
+  commandParam: string | vscode.Uri;
 }
 
-export async function resolveRequirements(): Promise<RequirementsData> {
-  const javaHome = await checkJavaRuntime();
+export async function resolveRequirements(context: vscode.ExtensionContext): Promise<RequirementsData> {
```
mathieu-suen-sonarsource: If you flag you function as `async` there is no point returning a `Promise` (It would de-sugare as Promise<Promise<...>>)

src/requirements.ts
```diff
@@ -145,3 +170,39 @@ function invalidJavaHome(reject, cause: string) {
     });
   }
 }
+
+export async function installManagedJre() {
+  return vscode.window.withProgress({ location: vscode.ProgressLocation.Notification, title: 'Downloading JRE' },
+    async (progress, cancelToken) => {
+      const options = await PlatformInformation.GetPlatformInformation();
+      console.log(`Detected options ${JSON.stringify(options)}`);
+      return jre.download({
```
mathieu-suen-sonarsource: You are mixing async/await then/catch pattern in the same function here I would rather chose one or the other.

PR #74, opened 2020-02-03
SLVSCODE-32 Rework commands and showing rule description

henryju:


src/protocol.ts
```diff
@@ -0,0 +1,13 @@
+import { RequestType } from 'vscode-languageclient';
```
jblievremont: Copyright header is missing.

PR #77, opened 2020-02-04
SLVSCODE-98 Rework connected mode settings

henryju:


package.json
```diff
@@ -61,15 +61,43 @@
...
+                    "messages",
+                    "verbose"
+                  ],
+                  "default": "off"
+                },
+                "format": {
+                  "type": "string",
+                  "enum": [
+                    "text",
+                    "json"
```
henryju: Not part of the ticket, but I found it while looking at other extensions. It should allow to use https://github.com/microsoft/language-server-protocol-inspector.

package.json
```diff
@@ -127,6 +158,7 @@
           "scope": "window"
         },
         "sonarlint.connectedMode.servers": {
+          "deprecationMessage": "The setting is deprecated. Use `sonarlint.connectedMode.connections.sonarqube` or `sonarlint.connectedMode.connections.sonarcloud` instead.",
```
jblievremont: Nitpick: looks like markdown is not supported here (yet?)

package.json
```diff
@@ -153,33 +185,149 @@
...
+              "serverUrl": "https://<myServerUrl>",
+              "token": "<mySQToken>"
+            }
+          ],
+          "items": {
+            "type": "object",
+            "properties": {
+              "connectionId": {
+                "type": "string",
+                "description": "A unique identifier for this connection to be used as a reference in `#sonarlint.connectedMode.project#`. Only needed if you plan to use multiple connections to SonarQube/SonarCloud.",
```
jblievremont:
```suggestion
                "description": "A unique identifier for this connection to be used as a reference in `#sonarlint.connectedMode.project#`. Only needed if you plan to use multiple connections to SonarQube.",
```

package.json
```diff
@@ -153,33 +185,149 @@
...
+              "organizationKey": "<myOrgKey>",
+              "token": "<mySCToken>"
             }
+          ],
+          "items": {
+            "type": "object",
+            "properties": {
+              "connectionId": {
+                "type": "string",
+                "description": "A unique identifier for this connection to be used as a reference in `#sonarlint.connectedMode.project#`. Only needed if you plan to use multiple connections to SonarQube/SonarCloud.",
```
jblievremont:
```suggestion
                "description": "A unique identifier for this connection to be used as a reference in `#sonarlint.connectedMode.project#`. Only needed if you plan to use multiple connections to SonarCloud.",
```
It might be worth mentioning that several connections are needed to bind to different organizations.

package.json
```diff
@@ -153,33 +185,149 @@
...
+                "default": "default"
+              },
+              "organizationKey": {
+                "type": "string",
+                "description": "Only used for SonarCloud"
```
jblievremont: I think we can come up with a more accurate description now :smile:

package.json
henryju: I saw that, but backtips are not rendering too bad, and this is future proof. Or do you prefer to use  single quotes?

package.json
henryju: Here I was considering the case of people having a mix of SQ/SC connections. The id should be unique across all kind of connections.

package.json
henryju: See my previous comment about SQ/SC. I'll add something about mutliple orgs.

package.json
henryju: :+1:

package.json
jblievremont: Nope, fine by me!

PR #78, opened 2020-02-06
SLVSCODE-116 Avoid early start of standalone engine

jblievremont:


src/rules.ts
```diff
@@ -93,11 +94,18 @@ export class AllRulesTreeDataProvider implements VSCode.TreeDataProvider<AllRule
       });
   }
 
+  private getAllRules(): Thenable<RulesResponse> {
+    if (! this.allRules) {
```
henryju: Extra space?

PR #79, opened 2020-02-07
SLVSCODE-8 Fix the verbose log switch on client side

henryju:


src/extension.ts
```diff
@@ -102,8 +102,8 @@ function logWithPrefix(data, prefix) {
   }
 }
 
-function isVerboseEnabled() {
-  return currentConfig.get('sonarlint.output.showVerboseLogs');
+function isVerboseEnabled(): boolean {
+  return currentConfig.get('output.showVerboseLogs', false);
```
henryju: The bug was here (wrong setting key)

PR #90, opened 2020-05-22
SLVSCODE-129 update java analyzer version to 6.4

kirill-knize-sonarsource:

### DIFF
--- a/its/src/test/javaSuite/java.test.ts
+++ b/its/src/test/javaSuite/java.test.ts
@@ -14,6 +14,7 @@ import * as vscode from 'vscode';
 // import * as myExtension from '../extension';
 
 const sampleFolderLocation = '../../../samples/';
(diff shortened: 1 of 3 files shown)


its/src/test/javaSuite/java.test.ts
```diff
@@ -14,6 +14,7 @@ import * as vscode from 'vscode';
 // import * as myExtension from '../extension';
 
 const sampleFolderLocation = '../../../samples/';
+const sampleJavaFolderLocation = '../../../samples/sample-java-maven-multi-module/';
```
henryju, line 17: Why duplicating almost the same constant than the previous one? You could keep this one and use it as well line 33

kirill-knize-sonarsource: Fixed.

PR #104, opened 2020-07-15
SLVSCODE-134 Adapt to the LightWeight mode of the Java LS

jblievremont:


its/src/test/runTest.ts
```diff
@@ -49,7 +49,23 @@ async function main() {
...
+
+		await runTests({
+			// Use the specified `code` executable
+			vscodeExecutablePath,
+			extensionDevelopmentPath,
+			extensionTestsPath: javaExtensionTestsPath,
+			launchArgs: [javaTestWorkspace]
+		});
+
+} catch (err) {
```
henryju: Indentation issues

src/java.ts
```diff
@@ -36,6 +48,29 @@ export function installClasspathListener(languageClient: SonarLintExtendedLangua
...
 
+export function installServerModeListener() {
+  const extension: VSCode.Extension<any> | undefined = getJavaExtension();
+  if (extension?.isActive) {
+    if (!serverModeListener) {
+      const extensionApi: any = extension.exports;
+      if (extensionApi && isJavaApiRecentEnough(extensionApi.apiVersion) && extensionApi.onDidServerModeChange) {
+        const onDidServerModeChange: VSCode.Event<ServerMode> = extensionApi.onDidServerModeChange;
+        serverModeListener = onDidServerModeChange(function(serverMode) {
+          if(serverMode !== ServerMode.LIGHTWEIGHT) {
```
henryju, line 59: Formatting issue.

src/java.ts
```diff
@@ -51,11 +86,14 @@ export async function getJavaConfig(
   languageClient: SonarLintExtendedLanguageClient,
   fileUri: string
 ): Promise<GetJavaConfigResponse> {
-  const extension: VSCode.Extension<any> | undefined = VSCode.extensions.getExtension('redhat.java');
+  const extension: VSCode.Extension<any> | undefined = getJavaExtension();
   try {
     const extensionApi: any = await extension?.activate();
     if (extensionApi && isJavaApiRecentEnough(extensionApi.apiVersion)) {
       installClasspathListener(languageClient);
+      if (extensionApi.serverMode === ServerMode.LIGHTWEIGHT) {
```
henryju, line 97: We could have a more defensive approach by checking if mode is either 'standard' or 'hybrid', and log that we don't support any other mode. I'm afraid the Java LS adding/renaming modes in the future.

src/java.ts
```diff
@@ -89,8 +90,10 @@ export async function getJavaConfig(
   const extension: VSCode.Extension<any> | undefined = getJavaExtension();
   try {
     const extensionApi: any = await extension?.activate();
+    console.log(Object.keys(extensionApi));
```
henryju, line 93: Do we still need this?

its/src/test/runTest.ts
```diff
@@ -41,32 +41,46 @@ async function main() {
...
+				extensionDevelopmentPath,
+				extensionTestsPath: javaExtensionTestsPath,
+				launchArgs: [javaTestWorkspace]
+			});
+		} catch(testError) {
+			testErrors.push(testError);
+		}
+
+		if (testErrors.length > 0) {
+			throw new Error('At least one test suite failed, please check logs above for actual failure.');
```
henryju, line 79: I wonder if you should not simply rethrow `testErrors[0]` to have a better stack trace in case the error is in the test script itself?

src/java.ts
jblievremont: Good catch :+1:

its/src/test/runTest.ts
jblievremont: Actually, I think we can be a bit more clever and provide a nicer developer UX:
- if there is a single error, re-throw it as it is
- if there are several errors, re-throw one with a message that aggregates the other messages

its/userdir/User/settings.json
```diff
@@ -0,0 +1,14 @@
+{
+  // Start with no editor open
+  "workbench.startupEditor": "none",
+  // Disable telemetry pings to Microsoft
+  "telemetry.enableTelemetry": false,
```
henryju, line 5: :+1:

PR #107, opened 2020-07-22
SLVSCODE-141 Update screen captures with Python example in recent VS Code

jblievremont:


henryju:
@christophelevis I think this can be improved in a next sprint.

PR #116, opened 2020-09-21
SLVSCODE-132 Move LS JRE path property out of settings synchronization

jblievremont:


henryju (approved):
LGTM, not manually tested.

PR #118, opened 2020-09-25
Kk/slvscode 138

kirill-knize-sonarsource:


jblievremont:
Rebased and merged as part of #119

PR #122, opened 2020-10-15
Disable telemetry when running ITs

jblievremont:


henryju (approved):
> [Info  - 14:56:54.387] Telemetry disabled by system property

in logs !

PR #126, opened 2020-12-10
SLVSCODE-113 Show developer notifications

jblievremont:


package.json
```diff
@@ -226,6 +226,10 @@
               "token": {
                 "type": "string",
                 "description": "Token generated from My Account>Security in SonarQube"
+              },
+              "disableNotifications": {
+                "type": "boolean",
+                "description": "Do not receive notifications from this server"
```
henryju, line 232: Maybe try to be more explicit, like we did in SLI. Maybe something like: 
> SonarLint will show notifications for bound projects when there are new issues assigned to you, or when the quality gate status changed. Set to false to disable all notifications for projects bound to this connection.

package.json
```diff
@@ -265,6 +269,10 @@
               "token": {
                 "type": "string",
                 "description": "Token generated from [My Account>Security](https://sonarcloud.io/account/security/) in SonarCloud"
+              },
+              "disableNotifications": {
+                "type": "boolean",
+                "description": "Do not receive notifications for this organization"
```
henryju, line 275: Same, what about:
> SonarLint will show notifications for bound projects when there are new issues assigned to you, or when the quality gate status changed. Set to false to disable all notifications for projects bound to this connection.

henryju (approved):

henryju:
Use case that doesn't work:
* have a connection with id `sonarcloud` and a folder binding to connection `sonarcloud`
* in folder settings, change connection id to `sonarcloud2`
* In console, we can see:
> De-registering notifications for org.sonarsource.clirr:sonar-clirr-plugin on sonarcloud
* now go the global settings and change connection id to `sonarcloud2` as well
Nothing occurs, while I would have expected notifications for be registered again.

Changing first the global settings and then folder settings works fine.

henryju:
Tested on Windows, with SonarCloud!

PR #127, opened 2020-12-15
SLVSCODE-113 Add link to connection settings in notification

jblievremont:


src/extension.ts
```diff
@@ -357,6 +356,12 @@ function installCustomRequestHandlers(context: VSCode.ExtensionContext) {
   languageClient.onRequest(protocol.BrowseTo.type,
     browseTo => VSCode.commands.executeCommand(Commands.OPEN_BROWSER, VSCode.Uri.parse(browseTo))
   );
+  languageClient.onRequest(protocol.OpenConnectionSettings.type,
+    isSonarCloud => {
+      const targetSection = `connectedMode.connections.${isSonarCloud ? 'sonarcloud' : 'sonarqube'}`;
```
henryju, line 361: Not tested, but why not:
```suggestion
      const targetSection = `sonarlint.connectedMode.connections.${isSonarCloud ? 'sonarcloud' : 'sonarqube'}`;
```

jblievremont: Ah good catch! It does work (there is a very little chance that another extension contributes a key like this), but it's a bad leftover of a previous iteration and it should be cleaner with the `sonarlint` prefix.

PR #131, opened 2021-01-15
SLVSCODE-164 Show security hotspot as clent-side diagnostic

jblievremont:


jblievremont:
Rebased and merged as part of #133

PR #133, opened 2021-01-27
SLVSCODE-165 Show hotspot description

jblievremont:


src/hotspots.ts
```diff
@@ -0,0 +1,142 @@
...
+  .then(action => {
+      if (action === 'Show Documentation') {
+        vscode.commands.executeCommand(
+          Commands.OPEN_BROWSER,
+          vscode.Uri.parse('https://docs.sonarqube.org/latest/user-guide/security-hotspots/')
+        );
+      }
+    });
+  } else if (foundUris.length > 1) {
+    // TODO Show quick pick to allow user to select the "right" file
```
henryju, line 37: What about taking the first candidate? In case of duplicate filenames, there is a possibility of duplicate content as well, so this is not very important which one is opened.

src/hotspots.ts
```diff
@@ -0,0 +1,142 @@
...
+export const hotspotsCollection = vscode.languages.createDiagnosticCollection('sonarlint-hotspots');
+
+let activeHotspot: RemoteHotspot;
+
+export const showSecurityHotspot = async (hotspot: RemoteHotspot) => {
+  const foundUris = await vscode.workspace.findFiles(`**/${hotspot.filePath}`);
+  if (foundUris.length === 0) {
+    vscode.window.showErrorMessage(`Could not find file '${hotspot.filePath}' in the current workspace.
+
+Please make sure that the right folder is open and bound to the right project on the server,
```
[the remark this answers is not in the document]
jblievremont: I'm not comfortable with using the term "project": AFAICT in VSCode the associated concept is "folder" (e.g "Open Folder", "Workspace Folders" etc.)

And for the [SQ|SC] part, I'm tempted to push this to a later (hardening?) sprint, once the feature is available on SonarCloud.

src/hotspots.ts
jblievremont: IMHO this case has a very low probability of happening, so :+1: for using the first one.

src/hotspots.ts
```diff
@@ -0,0 +1,145 @@
...
+          Commands.OPEN_BROWSER,
+          vscode.Uri.parse('https://docs.sonarqube.org/latest/user-guide/security-hotspots/')
+        );
+      }
+    });
+  } else {
+    activeHotspot = hotspot;
+    const documentUri = foundUris[0];
+    if (foundUris.length > 1) {
+      logToSonarLintOutput(`Multiple candidates found for '${hotspot.filePath}', using first match '${documentUri}'`);
```
henryju, line 42: :+1:

henryju (approved):

PR #157, opened 2021-07-23
SLVSCODE-187 vcs check

kirill-knize-sonarsource:


src/extension.ts
```diff
@@ -407,6 +408,53 @@ function installCustomRequestHandlers(context: VSCode.ExtensionContext) {
...
+
+  if (serr) {
+    throw new Error(serr);
+  }
+
+  return Promise.resolve(sout.length > 0);
+}
+
+async function isIgnoredByScm(fileUri: string): Promise<boolean> {
+  const uriRelative = VSCode.workspace.asRelativePath(fileUri, false);
```
henryju, line 435: The doc of this method says it will return the file itself if there is no workspace folders or if the file is not part of any of them.

src/extension.ts
```diff
@@ -407,6 +408,53 @@ function installCustomRequestHandlers(context: VSCode.ExtensionContext) {
...
+  const command = `git check-ignore ${uriRelative}`;
```
henryju, line 436: So if you open a workspace folder, and then open a file that is outside of it, it will fail with this error:
```
[Debug - 18:01:27.490] Queuing analysis of file 'file:///home/julien/Prog/ignored.js'
Error on git gitCommand: Error: Command failed: git check-ignore /home/julien/Prog/ignored.js
fatal: /home/julien/Prog/ignored.js : '/home/julien/Prog/ignored.js' est hors du dépôt à '/home/julien/Prog/Projects/sonarlint/sonarlint-vscode'

[stderr] Exception in thread "SonarLint analysis" java.util.concurrent.CompletionException: org.eclipse.lsp4j.jsonrpc.ResponseErrorException: Request sonarlint/isIgnoredByScm failed with message: Command failed: git check-ignore /home/julien/Prog/ignored.js
[stderr] fatal: /home/julien/Prog/ignored.js : '/home/julien/Prog/ignored.js' est hors du dépôt à '/home/julien/Prog/Projects/sonarlint/sonarlint-vscode'
[stderr] 	at java.base/java.util.concurrent.CompletableFuture.reportJoin(CompletableFuture.java:412)
[stderr] 	at java.base/java.util.concurrent.CompletableFuture.join(CompletableFuture.java:2044)
[stderr] 	at org.sonarsource.sonarlint.ls.AnalysisManager.lambda$analyze$5(AnalysisManager.java:307)
[stderr] 	at java.base/java.util.HashMap.computeIfAbsent(HashMap.java:1133)
[stderr] 	at org.sonarsource.sonarlint.ls.AnalysisManager.analyze(AnalysisManager.java:307)
[stderr] 	at org.sonarsource.sonarlint.ls.AnalysisManager.lambda$analyzeAsync$4(AnalysisManager.java:297)
[stderr] 	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
[stderr] 	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
[stderr] 	at java.base/java.lang.Thread.run(Thread.java:829)
```

src/extension.ts
```diff
@@ -407,6 +408,53 @@ function installCustomRequestHandlers(context: VSCode.ExtensionContext) {
...
+  let ignoredForAtLeastOneWorkspaceFolder = false;
+  const workspaceFolders = VSCode.workspace.workspaceFolders;
+  if (!isIterable(workspaceFolders)) {
+    return false;
+  }
+  for (const workspaceFolder of workspaceFolders) {
```
henryju, line 442: To me the algorithm is wrong, and can easily consider a file as ignored while it is not. Say you have 2 workspace folders: folderA and folderB.
folderA/foo.js is ignored
folderB/foo.js is not ignored
When trying to analyze folderB/foo.js, uriRelative will be equal to foo.js. Then you will check for each workspace folder if foo.js is ignored, which is true for folderB. So you'll skip the analysis...
It would be better to first decode from which workspace folder the file belongs to, and then only check the ignore in this folder.

src/extension.ts
```diff
@@ -407,6 +408,53 @@ function installCustomRequestHandlers(context: VSCode.ExtensionContext) {
   languageClient.onRequest(protocol.ShowTaintVulnerabilityRequest.type, showAllLocations);
 }
 
+async function isFileIgnoredForFolder(workspaceFolderPath: string, gitCommand: string): Promise<boolean> {
+  const {sout, serr} = await new Promise<{ sout: string, serr: string }>((resolve, reject) => {
+    ChildProcess.exec(
+        gitCommand,
```
henryju, line 414: What is happening if git CLI is not installed on the user machine? Or not in the PATH (quite possible on Windows).

src/extension.ts
henryju: > est hors du dépôt à

means

> is outside repository at

henryju (changes requested):
The added code is very complex, I would isolate it in a separate file, and try to have a good coverage with unit tests/integration tests, since I fear we might have a lot of troubles to stabilize it.
Works for simple case, but I quickly found a way to make it crash :(
IMO this is the tricky part of the sprint, so it should be polished.

src/extension.ts
```diff
@@ -431,28 +432,28 @@ async function isFileIgnoredForFolder(workspaceFolderPath: string, gitCommand: s
...
 
-function isIterable(obj): boolean {
-  if (obj == null) {
-    return false;
+export async function performIsIgnoredCheck(fileUri: string,
+                                            scmCheck: (workspaceFolderPath: string, gitCommand: string) =>
+                                                Promise<boolean>): Promise<boolean> {
+  const parsedFileUri = VSCode.Uri.parse(fileUri);
+  const workspaceFolder = VSCode.workspace.getWorkspaceFolder(parsedFileUri);
+  if(workspaceFolder == null) {
```
henryju, line 445: Code format issue. I don't remember exactly, but you should probably install prettier extension in VSCode (@jblievremont do you remember)

src/git.d.ts
```diff
@@ -0,0 +1,327 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
```
henryju, line 3: Shouldn't we update this header? The license.txt will not be in the project root.

src/git.d.ts
```diff
@@ -0,0 +1,327 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import { Uri, Event, Disposable, ProviderResult } from 'vscode';
```
henryju, line 6: Do we need to copy all those interfaces? Shouldn't we simply copy the one we use?

henryju (commented):
Looks ok to me (I would still like to see an end to end test (=IT)), but it can be added later. I would like to have @jblievremont do a review as well before merging.

henryju (approved):

PR #160, opened 2021-08-02
SLVSCODE-187 first secret detection notification

kirill-knize-sonarsource:

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -224,6 +224,10 @@ function toggleRule(level: ConfigLevel) {
   };
 }
 
+function openProblemsView() {
+  VSCode.commands.executeCommand('workbench.panel.markers.view.focus');
+}
+
 export function activate(context: VSCode.ExtensionContext) {
   currentConfig = getSonarLintConfiguration();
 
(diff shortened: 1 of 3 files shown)


src/extension.ts
```diff
@@ -359,6 +361,13 @@ export function activate(context: VSCode.ExtensionContext) {
   installClasspathListener(languageClient);
 }
 
+async function showNotificationForFirstSecretsIssue(context: VSCode.ExtensionContext) {
+  VSCode.window.showWarningMessage("SonarLint detected some secrets in one of the open files.\n" +
+      "We strongly advise you to review those secrets and ensure they are not committed into repositories. " +
+      "Please refer to the SonarLint tool window for more information.");
```
jblievremont, line 367: There is no SonarLint tool window in VSCode. Would it be possible to add an action to focus on the "Problems" view instead?

jblievremont (changes requested):

src/extension.ts
kirill-knize-sonarsource: Done.

jblievremont (approved):
LGTM 👍🏻

PR #161, opened 2021-08-03
SLVSCODE-187 make private analyzers downloading optional

kirill-knize-sonarsource:


scripts/prepare.js
```diff
@@ -46,22 +51,30 @@ function downloadIfNeeded(url, dest) {
...
+      request(url + '.sha1', callback);
+    }
   }
 }
 
 function downloadIfChecksumMismatch(expectedChecksum, url, dest) {
   if (!fs.existsSync(dest)) {
-    request(url, { auth }).pipe(fs.createWriteStream(dest));
-
+    if (credentialsDefined) {
```
jblievremont, line 73: This check also needs to be performed on line 85, otherwise the download of JAR files fails when credentials are not defined.

jblievremont (changes requested):

scripts/prepare.js
```diff
@@ -46,30 +52,38 @@ function downloadIfNeeded(url, dest) {
...
+      request(url, { auth }).pipe(fs.createWriteStream(dest));
+    } else {
+      request(url).pipe(fs.createWriteStream(dest));
+    }
```
jblievremont, line 78:
```suggestion
    sendRequest(url).pipe(fs.createWriteStream(dest));
```

kirill-knize-sonarsource: Done.

jblievremont (approved):

PR #168, opened 2021-09-06
SLVSCODE-197 Persist telemetry in context-provided global storage

jblievremont:


src/extension.ts
```diff
@@ -241,7 +241,7 @@ export function activate(context: VSCode.ExtensionContext) {
     initializationOptions: () => {
       return {
         productKey: 'vscode',
-        telemetryStorage: Path.resolve(context.extensionPath, '..', 'sonarlint_usage'),
+        telemetryStorage: Path.resolve(context.globalStorageUri.fsPath, 'telemetry', 'usage'),
```
henryju, line 244: Do we really need this `usage` subfolder? I'm tempted to use the smallest possible path to avoid issues on Windows.

henryju (changes requested):

src/extension.ts
jblievremont: This is the full path to the file. We could merge `telemetry` and `usage` into `telemetry_usage` for instance, which would result in:
- `C:\Users\username\AppData\Roaming\Code\User\globalStorage\sonarsource.sonarlint-vscode\telemetry_usage` on Windows
- `$HOME/.config/Code/User/globalStorage/sonarsource.sonarlint-vscode/telemetry_usage` on Linux
- Probably `$HOME/Library/Application Support/Code/User/globalStorage/sonarsource.sonarlint-vscode/telemetry_usage` on macOS

henryju: Oh, my bad, I though it was a folder. That's good like that, let's keep a telemetry folder in case we want to store more telemetry related things later.

henryju (approved):

PR #187, opened 2022-01-14
Try to troubleshoot issue with test

jblievremont:


jblievremont:
Could not understand why these tests fail while the `initScm` function (invoked later) is OK.

PR #190, opened 2022-01-20
SLVSCODE-222 Make dependency on git explicit in extension manifest

jblievremont:


henryju:
Are we sure about this move? Shouldn't we make the dependency optional (like we do for vscode-java)? What if a user doesn't use Git at all?

nixel2007:
Or svn...

jblievremont:
@henryju you're right, it should not be a hard dependency. Not merging this, and marking the ticket as won't fix with the explanation for documentation.

PR #208, opened 2022-03-18
SLVSCODE-258 Update embedded CFamily analyzer to 6.32.0.44918

mpaladin:


mpaladin:
@henryju this version removes the use of legacy keys in CFamily plugin, is there any logic in `sonarlint-vscode` that should be modified?

henryju:
> @henryju this version removes the use of legacy keys in CFamily plugin, is there any logic in `sonarlint-vscode` that should be modified?

No, we don't have any code relying on rule keys. And since we have not yet released with C++ suport, we don't care about settings migration.

henryju:
@abbas-sabra-sonarsource except if SonarLint is smart enough to translate deprecated keys from the server to new keys for the local analyzer. This is maybe already implemented, it is in my TODO to verify.

PR #225, opened 2022-04-27
SLVSCODE-240 Automate deployment to VS marketplace

jblievremont:


tomverin (approved):

damien-urruty-sonarsource (approved):
I had a doubt if a file downloaded in one step could be used in a different step, apparently it should be OK. LGTM :+1:

PR #228, opened 2022-05-03
SLVSCODE-276 Check files are open in an editor before triggering analysis

jblievremont:
See [matching changes on LS](https://github.com/SonarSource/sonarlint-language-server/pull/133)


src/extension.ts
```diff
@@ -582,6 +583,10 @@ export async function performIsIgnoredCheck(
   }
 }
 
+function isOpenInEditor(fileUri: string) {
+  return (VSCode.workspace.textDocuments.findIndex(document => document.uri.toString(true) === fileUri) >= 0);
```
damien-urruty-sonarsource, line 587: Minor improvement: I think we could use `some` instead of `findIndex`. Saves a few chars :smile:

damien-urruty-sonarsource (approved):
:+1:

src/extension.ts
jblievremont: Good catch, thanks! 👍🏻

PR #236, opened 2022-05-30
SLVSCODE-292 Fix issue with URI-encoded characters in file paths

jblievremont:


henryju (approved):
LGTM but I think we could ask users who reported the issue to validate the fix before releasing.

PR #238, opened 2022-06-07
SLVSCODE-294 New SonarQube connection UI + command

jblievremont:


src/connectionsetup.ts
```diff
@@ -0,0 +1,153 @@
...
+import * as util from './util';
+
+let connectionSetupPanel: vscode.WebviewPanel;
+
+export function connectToSonarQube(context: vscode.ExtensionContext) {
+  return () => {
+    lazyCreateConnectionSetupPanel(context);
+    connectionSetupPanel.webview.html = renderConnectionSetupPanel(context, connectionSetupPanel.webview);
+    connectionSetupPanel.webview.onDidReceiveMessage(handleMessage);
+    connectionSetupPanel.iconPath = {
```
sophio-japharidze-sonarsource, line 22: From WebViewPanel - iconPath?: Uri | { light: Uri; dark: Uri };
so you can only specify `util.resolveExtensionFile('images/sonarqube.svg')` since values for both `light` and `dark` are the same

src/connectionsetup.ts
```diff
@@ -0,0 +1,153 @@
...
+  const toolkitUri = resolver.resolve(
+    'node_modules',
+    '@vscode',
+    'webview-ui-toolkit',
+    'dist',
+    'toolkit.js'
+  );
+  const webviewMainUri = resolver.resolve('webview-ui', 'connectionsetup.js');
+
+  return `<!doctype html><html lang="en">
```
sophio-japharidze-sonarsource, line 62: just an idea, can/should this html snippet be moved to a separate file?

src/connectionsetup.ts
```diff
@@ -0,0 +1,153 @@
...
+/*
+ * Exported for unit tests
+ */
+export async function handleMessage(message) {
+  switch(message.command) {
+    case 'openTokenGenerationPage':
+      await openTokenGenerationPage(message);
+      break;
+    case 'saveConnection':
+      delete message.command;
```
sophio-japharidze-sonarsource, line 108: Just curious, what is happening here?

src/connectionsetup.ts
jblievremont: It could, as long as it remains static; my idea is to make it dynamic (soon) to support SonarCloud, and reuse it for the modification of a connection.

src/connectionsetup.ts
jblievremont: I'm reusing the passed message "as is" to save it to the settings (since it already has the right members), just removing the `command` attribute 🙂

jblievremont:
Closing this PR, see #239 for the new one that has a cleaner branch name and targets the sprint branch 🙂

PR #239, opened 2022-06-09
SLVSCODE-294 Add command and UI to create a SonarQube connection

jblievremont:


jblievremont:
IMHO it does not make a lot of sense to link these changes to a particular ticket/MMF, it's more like preparatory work. I'll make sure that they stay in their own commit when I squash/rebase before merging.

PR #240, opened 2022-06-15
SLVSCODE-295 Add a command to edit a connection given its ID

jblievremont:


package.json
```diff
@@ -452,6 +453,11 @@
         "command": "SonarLint.ConnectToSonarQube",
         "title": "Connect To SonarQube",
         "category": "SonarLint"
+      },
+      {
+        "command": "SonarLint.ModifyConnection",
```
sophio-japharidze-sonarsource, line 458: So will SonarQube and SonarCloud modification use the same command? Just asking as in my branch I have two separate commands `SonarLint.EditSonarQubeConnection` and `SonarLint.EditSonarCloudConnection`

src/extension.ts
```diff
@@ -338,6 +338,9 @@ export function activate(context: VSCode.ExtensionContext) {
   context.subscriptions.push(
     VSCode.commands.registerCommand(Commands.CONNECT_TO_SONARQUBE, connectToSonarQube(context))
   );
+  context.subscriptions.push(
+      VSCode.commands.registerCommand(Commands.MODIFY_CONNECTION, editConnection(context))
```
sophio-japharidze-sonarsource, line 342: Minor thing but maybe it would be better to use uniform terminology, either `edit` or `modify` or `update`

webview-ui/connectionsetup.js
```diff
@@ -29,17 +29,40 @@ function init() {
...
+
+function hasValidServerUrl() {
+  /**
+   * @type {HTMLInputElement}
+   */
+  const serverUrlInput = byId('serverUrl');
+  return serverUrlInput.validity.valid && isValidUrl(serverUrlInput.value);
+}
+
+function isValidUrl(value) {
```
sophio-japharidze-sonarsource, line 59: I think it would be great to have tests for this method

sophio-japharidze-sonarsource (commented):
Looks good. Added comments/questions for some smaller details.

src/connectionsetup.ts
```diff
@@ -15,19 +15,49 @@ import { ResourceResolver } from './webview';
...
 
-export function reportConnectionCheckResult(result: ConnectionCheckResult) {
+export function editConnection(context: vscode.ExtensionContext) {
+  return async (connectionId: string) => {
+    if (! connectionId) {
+      // TODO This selection step should be removed once hooked up to the tree view
+      const connectionIds = vscode.workspace.getConfiguration(SONARLINT_SETTINGS_KEY)
+          .get<Array<SonarQubeConnection>>(SONARQUBE_CONNECTIONS_KEY)
+          .map(c => c.connectionId);
+      connectionId = await vscode.window.showQuickPick(connectionIds);
```
sophio-japharidze-sonarsource, line 44: What if there is only one connection without connectionId? the `default` one

package.json
jblievremont: I'm not sure, maybe it would be cleaner to have separate commands (even if they load the same webview ultimately). I'll make sure to use `SonarLint.EditSonarQubeConnection` here to make the merge more straightforward.

webview-ui/connectionsetup.js
jblievremont: This might prove tricky: this file is only used in the webview and is completely separated from the TS module system.

sophio-japharidze-sonarsource: ack

src/connectionsetup.ts
```diff
@@ -34,14 +34,18 @@ export function connectToSonarQube(context: vscode.ExtensionContext) {
...
-export function editConnection(context: vscode.ExtensionContext) {
+export function editSonarQubeConnection(context: vscode.ExtensionContext) {
   return async (connectionId: string) => {
     if (! connectionId) {
       // TODO This selection step should be removed once hooked up to the tree view
       const connectionIds = vscode.workspace.getConfiguration(SONARLINT_SETTINGS_KEY)
           .get<Array<SonarQubeConnection>>(SONARQUBE_CONNECTIONS_KEY)
           .map(c => c.connectionId);
-      connectionId = await vscode.window.showQuickPick(connectionIds);
+      if (connectionId.length === 1) {
```
sophio-japharidze-sonarsource, line 44:
```suggestion
      if (connectionIds.length === 1) {
```

sophio-japharidze-sonarsource: I think you meant this?

PR #241, opened 2022-06-16
SLVSCODE-100 secure storage for token

kirill-knize-sonarsource:


src/settings.ts
```diff
@@ -0,0 +1,104 @@
...
+    if (!await secureStorage.hasTokenForServer(connection.serverUrl)) {
+      return true;
+    }
+  }
+  return false;
+}
+
+export function getSonarLintConfiguration(secureStorage?: SonarLintSecretStorage): VSCode.WorkspaceConfiguration {
+  const settingsFromConfigFiles = getSettingsFileConfig();
+  (async () => migrateSettings(settingsFromConfigFiles, secureStorage))();
```
sophio-japharidze-sonarsource, line 24: I think if you don't need to `await` response you can directly just call this function without `async` wrapper

package.json
```diff
@@ -146,8 +146,7 @@
             },
             "additionalProperties": false,
             "required": [
-              "serverUrl",
-              "token"
+              "serverUrl"
```
sophio-japharidze-sonarsource, line 149: What does this do?

src/settings.ts
sophio-japharidze-sonarsource: Also, are we migrating ALL configurations from file to secure storage? I thought it was only for connected mode settings

src/settings.ts
```diff
@@ -0,0 +1,104 @@
...
+  return settingsFromConfigFiles;
+}
+
+async function migrateSettings(settings: VSCode.WorkspaceConfiguration, secureStorage?: SonarLintSecretStorage) {
```
sophio-japharidze-sonarsource, line 28: I think this function should be called `migrateConnectedModeSettings` or similar,  to describe immediately that only connections are being migrated. But it's a minor detail

src/settings.ts
```diff
@@ -0,0 +1,104 @@
...
+    return this.secretStorage.get(serverId);
+  }
+
+  async hasTokenForServer(serverId: string): Promise<boolean> {
+    return (await this.getServerToken(serverId)) !== undefined;
+  }
+
+}
+
+type SqConnection = { serverUrl: string, token: string };
```
sophio-japharidze-sonarsource, line 104: It can also have `connectionId? : string`

package.json
kirill-knize-sonarsource: Since we migrating tokens from settings.json file to secret storage we don't ask for token as required field anymore.

src/settings.ts
kirill-knize-sonarsource: Changed to { serverId?: string, serverUrl?: string, organizationKey?: string, token?: string }

src/settings.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  async getServerToken(serverId: string): Promise<string | undefined> {
+    return this.secretStorage.get(serverId);
+  }
+
+  async hasTokenForServer(serverId: string): Promise<boolean> {
+    return (await this.getServerToken(serverId)) !== undefined;
+  }
+
+  async deleteTokenForServer(serverId: string): Promise<void> {
```
sophio-japharidze-sonarsource, line 86: What is `serverId` ? Does it refer to `connectionId` or `serverUrl` from current context?

kirill-knize-sonarsource: Yes.

PR #246, opened 2022-06-22
SLVSCODE-296 check connection for <default> in case id is undefined

sophio-japharidze-sonarsource:


jblievremont:
Merged with #241

PR #250, opened 2022-06-27
SLVSCODE-301 Add commands to create and update a SonarCloud connection

jblievremont:


src/connections.ts
```diff
@@ -57,12 +57,12 @@ export class AllConnectionsTreeDataProvider implements VSCode.TreeDataProvider<C
 
     async getConnections(type: string): Promise<Connection[]> {
         const contextValue = type === 'sonarqube' ? 'sonarqubeConnection' : 'sonarcloudConnection';
-        const labelKey = type === 'sonarqube' ? 'connectionId' : 'organizationKey';
```
sophio-japharidze-sonarsource, line 60: 👍

src/settings.ts
```diff
@@ -151,7 +162,44 @@ export class ConnectionSettingsService {
...
-  async addTokensFromSettingsToSecureStorage(sqConnections: SonarQubeConnection[]) {
+  async addSonarCloudConnection(connection: SonarCloudConnection) {
+    const connections = this.getSonarCloudConnections();
+    const newConnection: SonarCloudConnection = { organizationKey: connection.organizationKey };
+    if (connection.connectionId !== undefined) {
+      newConnection.connectionId = connection.connectionId;
+    }
+    if (connection.disableNotifications) {
+      newConnection.disableNotifications = true;
+    }
```
sophio-japharidze-sonarsource, line 173: equivalent to `newConnection.disableNotifications = connection.disableNotifications` ?

src/settings.ts
```diff
@@ -18,31 +18,42 @@ const SONARQUBE_CONNECTIONS_CATEGORY = `${SONARLINT_CATEGORY}.${CONNECTIONS_SECT
...
 async function hasUnmigratedConnections(sqConnections: SonarQubeConnection[],
+                                        scConnections: SonarCloudConnection[],
                                         settingsService: ConnectionSettingsService): Promise<boolean> {
   for (const connection of sqConnections) {
     if (!await settingsService.hasTokenForServer(connection.serverUrl) && connection.token) {
       return true;
     }
   }
+  for (const connection of scConnections) {
+    if (!await settingsService.hasTokenForServer(connection.organizationKey)) {
```
kirill-knize-sonarsource, line 29: Connection should have token, otherwise there is no point to migrate.

src/settings.ts
```diff
@@ -18,31 +18,42 @@ const SONARQUBE_CONNECTIONS_CATEGORY = `${SONARLINT_CATEGORY}.${CONNECTIONS_SECT
...
 
 async function hasUnmigratedConnections(sqConnections: SonarQubeConnection[],
+                                        scConnections: SonarCloudConnection[],
                                         settingsService: ConnectionSettingsService): Promise<boolean> {
   for (const connection of sqConnections) {
     if (!await settingsService.hasTokenForServer(connection.serverUrl) && connection.token) {
       return true;
     }
   }
+  for (const connection of scConnections) {
```
kirill-knize-sonarsource, line 28: May be we could extract this to method. Some duplication here.

src/settings.ts
```diff
@@ -151,7 +162,44 @@ export class ConnectionSettingsService {
...
+      throw new Error(`Could not find connection '${connection.connectionId}' to update`);
+    }
+    connectionToUpdate.organizationKey = connection.organizationKey;
+    if (connection.disableNotifications) {
+      connectionToUpdate.disableNotifications = true;
+    } else {
+      delete connectionToUpdate.disableNotifications;
+    }
+    await this.storeServerToken(connection.organizationKey, connection.token);
+    this.client.onTokenUpdate();
```
kirill-knize-sonarsource, line 193: We firing this event even if token didn't change. Maybe it's not critical.

kirill-knize-sonarsource (approved):
LGTM with comments.

package.json
```diff
@@ -151,13 +151,11 @@
             ],
             "examples": [
               {
-                "serverUrl": "https://<replace_with_your_sonarqube_server_url>",
```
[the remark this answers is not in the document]
jblievremont: Good point, and since the UI has issues in Codespaces, putting the token in the settings and letting the migration run is one possible workaround to configure connected mode.

src/settings.ts
jblievremont: Not exactly: we only want to add the `disableNotifications` attribute when it is `true` (`false` is the default value).

src/settings.ts
jblievremont: Yes thanks, I saw it after the rebase.

sophio-japharidze-sonarsource (approved):

PR #264, opened 2022-07-21
project binding service

kirill-knize-sonarsource:


kirill-knize-sonarsource:
Confirmed that workspace folder level settings overrides workspace and global levels.

src/binding.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  static init(): void {
+    BindingService._instance = new BindingService();
+  }
+
+  static get instance(): BindingService {
+    return BindingService._instance;
+  }
+
+  async updateWorkspaceFolderBinding(workspaceFolderUri: VSCode.Uri, projectKey: string): Promise<void> {
```
jblievremont, line 26: IMHO, this method should also accept an optional connection ID. With this signature, folders can only be bound to a project from the `<default>` connection.

I believe this module should export an interface (e.g `ProjectBinding`) to represent bindings. This method could then take one such object as parameter (and an eventual `getAllBindings` method could return a collection of those).

src/binding.ts
```diff
@@ -0,0 +1,36 @@
...
+    const config = VSCode.workspace.getConfiguration(SONARLINT_CATEGORY, workspaceFolderUri);
+    return config.update(PROJECT_BINDING_PROPERTY, { projectKey }, VSCode.ConfigurationTarget.WorkspaceFolder);
+  }
+
+  async deleteWorkspaceFolderBinding(workspaceFolderUri: VSCode.Uri): Promise<void> {
```
jblievremont, line 31: Could we remove the `WorkspaceFolder` part of the method names? It feels a bit redundant to me.

jblievremont (changes requested):
See my comments.

In theory, I would also like to see tests, although they might come later in the form of medium tests that call the commands.

src/binding.ts
```diff
@@ -0,0 +1,36 @@
...
+    BindingService._instance = new BindingService();
+  }
+
+  static get instance(): BindingService {
+    return BindingService._instance;
+  }
+
+  async updateWorkspaceFolderBinding(workspaceFolderUri: VSCode.Uri, projectKey: string): Promise<void> {
+    const config = VSCode.workspace.getConfiguration(SONARLINT_CATEGORY, workspaceFolderUri);
+    return config.update(PROJECT_BINDING_PROPERTY, { projectKey }, VSCode.ConfigurationTarget.WorkspaceFolder);
```
sophio-japharidze-sonarsource, line 28: I think an optional `connectionId` is also needed?

src/binding.ts
```diff
@@ -0,0 +1,64 @@
...
+    const config = VSCode.workspace.getConfiguration(SONARLINT_CATEGORY, workspaceFolderUri);
+    return config.update(PROJECT_BINDING_PROPERTY, undefined, VSCode.ConfigurationTarget.WorkspaceFolder);
+  }
+
+  async getAllBindings(): Promise<Map<string, Map<string, ProjectBinding[]>>> {
+    const bindings = new Map<string, Map<string, ProjectBinding[]>>();
+    for (const folder of VSCode.workspace.workspaceFolders || []) {
+      const config = VSCode.workspace.getConfiguration(SONARLINT_CATEGORY, folder.uri);
+      const binding = config.get<ProjectBinding>(PROJECT_BINDING_PROPERTY);
+      const projectKey = binding.projectKey;
```
kirill-knize-sonarsource, line 41: Here I'm checking against projectKey not for the binding itself because there is a possibility that user will have this in the settings:
```
    "sonarlint.connectedMode.project": {
    
    }
```
And it will be an existing binding but unlikely we want to display such things on frontend.

jblievremont: This is not a valid binding: a project key on the server cannot be null. So indeed, I believe it should not be shown in the UI.

jblievremont (approved):
LGTM 👍🏻 

I would love to see a few unit tests, although this will probably be tested by integration//medium tests a bit later.

sophio-japharidze-sonarsource (approved):

PR #265, opened 2022-07-21
SLVSCODE-319 Show remote projects and their bound workspace folders in connected mode view

jblievremont:


src/binding.ts
```diff
@@ -43,13 +43,13 @@ export class BindingService {
...
         if (!bindings.has(connectionId)) {
-          bindings.set(connectionId, new Map<string, ProjectBinding[]>());
+          bindings.set(connectionId, new Map());
         }
         connectionBindings = bindings.get(connectionId);
         if (!connectionBindings.has(projectKey)) {
           connectionBindings.set(projectKey, []);
         }
-        connectionBindings.get(projectKey).push(binding);
+        connectionBindings.get(projectKey).push({ folder, binding });
```
jblievremont, line 52: FYI @kirill-knize-sonarsource - I rebased my branch and updated the service to return the full `WorkspaceFolder` object alongside the binding; this way the folder's URI can be attached to the node in the tree (see below).

src/binding.ts
```diff
@@ -63,13 +64,13 @@ export class BindingService {
         const connectionId = binding.connectionId || binding.serverId || '<default>';
         let connectionBindings = bindings.get(connectionId);
         if (!bindings.has(connectionId)) {
```
sophio-japharidze-sonarsource, line 66: I think these two lines are useless? The `bindings` will always be empty here, as set on line 58.

src/binding.ts
```diff
@@ -63,13 +64,13 @@ export class BindingService {
...
-          bindings.set(connectionId, new Map<string, ProjectBinding[]>());
+          bindings.set(connectionId, new Map());
         }
         connectionBindings = bindings.get(connectionId);
         if (!connectionBindings.has(projectKey)) {
```
sophio-japharidze-sonarsource, line 70: I think this check is useless, it should always be true as set on line 67

src/binding.ts
jblievremont: It's a lazy initialization: initially, `bindings` is empty, and we are lazily creating a new `Map` for each `connectionId` encountered while iterating over the workspace folders (and same below, lazily creating an array to hold bindings for a given project the first time we find one folder bound to it).

I can maybe rename the variables to make it more explicit.

sophio-japharidze-sonarsource: Thanks for the clarification! I missed the for loop :)

sophio-japharidze-sonarsource (approved):
LGTM

PR #266, opened 2022-07-22
SLVSCODE-320 create binding ui

sophio-japharidze-sonarsource:


src/binding.ts
```diff
@@ -4,61 +4,195 @@
  * sonarlint@sonarsource.com
  * Licensed under the LGPLv3 License. See LICENSE.txt in the project root for license information.
  * ------------------------------------------------------------------------------------------ */
-'use strict';
```
jblievremont, line 7: I think we should keep it for now (and maybe switch to `alwaysStrict` in our tsconfig which would allow us to remove it from all source files).

src/binding.ts
```diff
@@ -4,61 +4,195 @@
...
 
 const SONARLINT_CATEGORY = 'sonarlint';
-const PROJECT_BINDING_PROPERTY = 'connectedMode.project';
+const BINDING_SETTINGS = 'connectedMode.project';
+const DEFAULT_CONNECTION_ID = '<default>';
+const OPEN_FOLDER_ACTION = 'Open Folder';
+const BIND_MANUALLY_ACTION = 'Bind Manually';
+
+export interface ProjectBinding {
+    connectionId?: string;
```
jblievremont, line 20: Indentation is 4-spaces instead of 2 (also in the test file) - maybe there is something missing in your IDE configuration?

src/binding.ts
```diff
@@ -4,61 +4,195 @@
...
+}
+
+function createSettingsFileIfNotExists(workspaceFolder: VSCode.WorkspaceFolder) {
+    if (!existsSync(workspaceFolder.uri.fsPath + '/.vscode')) {
+        mkdirSync(workspaceFolder.uri.fsPath + '/.vscode');
+    }
+    try {
+        writeFileSync((workspaceFolder.uri.fsPath + '/.vscode/settings.json'), '', { flag: 'wx' });
+    } catch {
+        return;
```
jblievremont, line 33: Maybe `logToSonarLintOutput` that something went wrong?

src/binding.ts
```diff
@@ -4,61 +4,195 @@
...
-      const binding = config.get<ProjectBinding>(PROJECT_BINDING_PROPERTY);
-      const projectKey = binding.projectKey;
-      if (projectKey) {
-        const connectionId = binding.connectionId || binding.serverId || '<default>';
-        let connectionBindings = bindings.get(connectionId);
-        if (!bindings.has(connectionId)) {
-          bindings.set(connectionId, new Map<string, ProjectBinding[]>());
+    private static _instance: BindingService;
+
+    static init(getRemoteProjectsForConnection: (connectionId: string) => Thenable<Map<string, string>>,
```
jblievremont, line 41: I wonder if we should pass the whole client object here (and mock it in the tests - see #265).

src/binding.ts
```diff
@@ -4,61 +4,195 @@
...
+                                VSCode.Uri.joinPath(workspaceFolder.uri, '/.vscode/settings.json'));
+                        }
+                    });
+            }
+
+            remoteProjects.forEach((v, k) => {
+                itemsList.push({
+                    label: v,
+                    description: k,
+                    buttons: [
```
jblievremont, line 184: 👍🏻

jblievremont (changes requested):
Looks good from a functional side, could you please reformat with 2-spaces soft tabs? I believe it would make the diff easier to read (and the source code more consistent).

PR #277, opened 2022-08-03
Add check for existing bindings list before deletion

sophio-japharidze-sonarsource:


jblievremont (changes requested):
I believe each change should be in its own commit, since they serve different purpose.

* The change in `src/binding.ts` can be linked to SLVSCODE-322
* The one on `scripts/dependencies.json` should mention that it's preparatory work for the release of 3.8
* The commit message for the changelog update can mention MMF-2839 and the 3.8 release

PR #291, opened 2022-09-05
SLVSCODE-326 Try to find analysis properties files in unbound workspace folders

jblievremont:


package.json
```diff
@@ -669,6 +669,7 @@
     "inly": "4.0.8",
     "luxon": "2.3.0",
     "path-exists": "4.0.0",
+    "properties": "1.2.1",
```
kirill-knize-sonarsource, line 672: Library is 9 years old and not being maintained. May we will need to think about better solution later.

kirill-knize-sonarsource (approved):

PR #292, opened 2022-09-05
SLVSCODE-330 ask for permission to autobind when many unbound folders

sophio-japharidze-sonarsource:


src/connected/autobinding.ts
```diff
@@ -0,0 +1,137 @@
...
+    }
+    const unboundFolders = VSCode.workspace.workspaceFolders.filter(workspaceFolder =>
+      this.bindingService.shouldBeAutoBound(workspaceFolder)
+    );
+    if (unboundFolders.length > AUTOBINDING_THRESHOLD) {
+      this.askUserBeforeAutoBinding();
+    }
+    unboundFolders.forEach(unboundFolder => {
+      this.autoBindFolder(unboundFolder);
+    });
```
jblievremont, line 54: I believe this should be in an `else` block: the way I understand the code, this path will be executed after the "too many folders path", regardless of whether there are indeed too many folders or not (confirmed by manual test).

jblievremont (changes requested):
See my comment on the "not too many folders" branch.

I'm also making a note that we will probably need separate workspace state flags for "do not ask again for this workspace" and "do not ask again for this folder".

jblievremont (approved):
LGTM 👍🏻!

PR #294, opened 2022-09-06
SLVSCODE-328 - Try to find a remote project to bind

kirill-knize-sonarsource:


src/connected/autobinding.ts
```diff
@@ -156,4 +165,161 @@ export class AutoBindingService {
...
+    const connectionToBestHits = new Map<BaseConnection, MatchHit[]>();
+    for (const [connection, projects] of connectionToServerProjects) {
+      let bestHits: MatchHit[] = [];
+      bestHits.push({ hits: 0, projectKey: '', connection: { connectionId: '' } });
+      for (const project of projects) {
+        const projectKey = project.key;
+        const projectName = project.name;
+        const serverProjectString = (projectKey + projectName).toLowerCase();
+        const folderNameHits = this.getHits(folderNameTokens, serverProjectString);
+        const workspaceNameHits = this.getHits(workspaceNameTokens, serverProjectString);
```
sophio-japharidze-sonarsource, line 185: Maybe we could optimise this by
- get hits for workspace
- if no hits, then get hits for folder
It could eliminate unnecessary processing

src/connected/autobinding.ts
```diff
@@ -156,4 +165,161 @@ export class AutoBindingService {
...
+      this.promptToAutoBindMultiChoice(connectionToBestHits, unboundFolder);
+    }
+  }
+
+  private async promptToAutoBindSingleOption(bestHit: MatchHit, connectionToBestHits: Map<BaseConnection,
+    MatchHit[]>, unboundFolder: VSCode.WorkspaceFolder) {
+    const [connection] = connectionToBestHits.keys();
+    const connectionName = getDisplayName(connection);
+    const result = await VSCode.window.showInformationMessage(
+      `There is a project to bind on ${connectionName}. Do you want to bind?`, 'Bind');
```
sophio-japharidze-sonarsource, line 215: I think it would be useful to add the matching server project name in the message. Something like
`There is a project ${bestHit.projectKey} on ${connectionName}. Do you want to bind?`

src/connected/autobinding.ts
```diff
@@ -156,4 +165,161 @@ export class AutoBindingService {
...
+    if (result === 'Bind') {
+      this.bindingService.saveBinding(bestHit.projectKey, connection.connectionId, unboundFolder);
+      VSCode.window.showInformationMessage(`Workspace folder '${unboundFolder.name}/'
+                      has been bound with project '${bestHit.projectKey}'`);
```
sophio-japharidze-sonarsource, line 219: Thought for refactoring - maybe we should move this info notification inside the `saveBinding()` method

src/connected/autobinding.ts
```diff
@@ -156,4 +165,161 @@ export class AutoBindingService {
...
+        const connection = Array.from(connectionToBestHits.keys()).find(c => c.connectionId === connectionId);
+        const serverType = getServerType(connection);
+        const baseServerUrl = this.bindingService.getBaseServerUrl(connection.connectionId, serverType);
+        remoteProjectsQuickPick.busy = true;
+        VSCode.commands.executeCommand(
+          Commands.OPEN_BROWSER,
+          VSCode.Uri.parse(`${baseServerUrl}?id=${projectKey}`)
+        );
+      });
+      remoteProjectsQuickPick.show();
```
sophio-japharidze-sonarsource, line 254: Let's maybetry during merge to directly call existing `createOrEditBinding()` method instead of duplicating the code

src/connected/autobinding.ts
kirill-knize-sonarsource: I don't think so. Because we need to match by both and choose best one.

jblievremont (approved):
LGTM 👍🏻

PR #297, opened 2022-09-13
SLVSCODE-340 Auto binding command

kirill-knize-sonarsource:


kirill-knize-sonarsource:
Cleaning of Don't ask me" flags should be added when they will be implemented.

sophio-japharidze-sonarsource (approved):

PR #304, opened 2022-09-26
SLVSCODE-350 Open new token generation page in SQ 9.7+

jblievremont:


src/lsp/protocol.ts
```diff
@@ -258,4 +258,15 @@ export namespace GetRemoteProjectNames {
   );
 }
 
+export interface TokenGenerationParams {
+  serverPath?: string;
+  errorMessage?: string;
+}
+
+export namespace GetTokenGenerationParams {
+  export const type = new lsp.RequestType<string, TokenGenerationParams, null>(
```
sophio-japharidze-sonarsource, line 267: Minor - why `TokenGenerationParams`? Isn't the request for `GetTokenGenerationUrl`

src/lsp/client.ts
```diff
@@ -46,4 +46,8 @@ export class SonarLintExtendedLanguageClient extends LanguageClient {
   getRemoteProjectsForConnection(connectionId: string) {
     return this.sendRequest(protocol.GetRemoteProjectsForConnection.type, { connectionId });
   }
+
+  getTokenGenerationParams(serverUrl: string): Promise<TokenGenerationParams> {
```
sophio-japharidze-sonarsource, line 50: Suggestion:
`getTokenGenerationUrl(baseServerUrl: string)....`

sophio-japharidze-sonarsource (approved):
LGTM with minor naming comments

src/lsp/protocol.ts
jblievremont: Yup good catch, this needs to be updated to the latest signature from the LS 👍🏻

PR #313, opened 2022-10-24
SLVSCODE-344 Show in rule description that SonarLint does not detect taint vulnerabilities in local code analysis

kirill-knize-sonarsource:


src/rules/rulepanel.ts
```diff
@@ -81,6 +86,20 @@ function base64encode(file: string) {
   return FS.readFileSync(file).toString('base64');
 }
 
+function renderTaintBanner(rule: ShowRuleDescriptionParams, infoImgSrc:string) {
+  if (!rule.isTaint) {
+    return '';
+  }
+  return `<div class="taint-banner-wrapper">
+            <p class="taint-banner"><span><img class="taint-info-icon" src=${infoImgSrc} alt="info"></span> 
+            This injection vulnerability was detected by the latest SonarQube or SonarCloud analysis.
```
jblievremont, line 95: Nitpick: I'm not a huge fan of the "SonarQube or SonarCloud" part, it should be possible to get the information from somewhere that the server that reported the issue is SonarCloud (or not).

jblievremont (approved):
Minor nitpick, else LGTM 👍🏻

src/rules/rulepanel.ts
sophio-japharidze-sonarsource: > it should be possible to get the information from somewhere that the server that reported the issue is SonarCloud (or not).

I think this is kind related to SLLS-106, now the taint issues will have a field `source` which will be specific to SQ/SC, so maybe it can be used here as well

PR #320, opened 2022-11-15
SLVSCODE-353 don't propose binding for failed connection

sophio-japharidze-sonarsource:


package.json
```diff
@@ -703,8 +703,8 @@
     "istanbul-lib-source-maps": "^3.0.6",
     "istanbul-reports": "^2.2.7",
     "map-stream": "^0.1.0",
-    "mocha": "9.1.3",
-    "mocha-multi-reporters": "^1.1.7",
+    "mocha": "10.1.0",
```
sophio-japharidze-sonarsource, line 706: Update to latest dependencies

src/connected/autobinding.ts
```diff
@@ -118,7 +118,10 @@ export class AutoBindingService {
       sqConnections
     );
     const connectionToBestHits = getBestHitsForConnections(connectionToServerProjects, unboundFolder);
-    this.promptToAutoBind(connectionToBestHits, unboundFolder);
+    if (connectionToBestHits.size > 0) {
+      // We don't want to show notification if it is not actionable
```
sophio-japharidze-sonarsource, line 122: There will be no matches if there is no successful connection + if there are really no matches

PR #334, opened 2023-01-24
SLVSCODE-381 Hotspots rule description: use tabbed description

kirill-knize-sonarsource:


jblievremont (changes requested):
From a pure frontend point of view, LGTM 👍🏻 

_And_ I think the new Language Server needs to be used in `scripts/dependencies.json` 😅

sophio-japharidze-sonarsource (approved):

PR #343, opened 2023-02-14
SLVSCODE-416 Use release sonar-text version

kirill-knize-sonarsource:


jblievremont (approved):
LGTM 👍🏻

Could you please make sure that the commit message mentions the actual version number before merging? I believe our future selves will thank you 😄

PR #348, opened 2023-02-23
DOC-488, use of Secrets with Connected Mode

kevin-hinz:


README.md
```diff
@@ -108,6 +108,9 @@ In SonarLint for VS Code 3.14 and above, local detection of [Security Hotspots](
 
 Please see the [documentation](https://github.com/SonarSource/sonarlint-vscode/wiki/Security-hotspots) for more details. 
 
+### Secrets detection
+Secrets are pieces of user-specific or system-level credentials that should be protected and accessible to legitimate users only. SonarLint now detects exposed Secrets in your source code and language agnostic config files when running in Connected Mode with SonarQube or SonarCloud.
```
sophio-japharidze-sonarsource, line 112: Not an expert opinion:
> SonarLint now detects
I would propose to 
> SonarLint detects
Because this is a README, and there is no clear reference of what `now` means. In changelog, it would be different for example.

sophio-japharidze-sonarsource: And, secrets have already been available in standalone mode for a long time. The change that happened now, it became available on SonarQube too, so we will honor Connected Mode settings for secrets too

README.md
```diff
@@ -120,6 +123,7 @@ Connected Mode will also allow unlocking of your analysis for these languages:
 
 - [Apex rules](https://rules.sonarsource.com/apex)
 - [PL/SQL rules](https://rules.sonarsource.com/plsql)
+- [Secrets rules](https://rules.sonarsource.com/secrets)
```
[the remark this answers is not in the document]
kevin-hinz: good point. It is actually the SQ/SC Quality Profile that is unlocked, not the analysis.

sophio-japharidze-sonarsource (approved):

PR #350, opened 2023-02-24
SLVSCODE-361 use auto binding algorithm from core

vojtech-suchy-sonarsource:
Update vscode to take advantage of [SonarLintCore Binding suggestions](https://github.com/SonarSource/sonarlint-core/blob/master/spec/binding_suggestion.adoc).


jblievremont (approved):
LGTM 👍🏻 it would be nice if we find a way to keep the "from connection 'Xxx'" bit in the notification message.

PR #351, opened 2023-02-24
SLVSCODE-414 Notify the language server of a branch change only when it actually changes

jblievremont:


vojtech-suchy-sonarsource (approved):
👍 Looks good.
(I haven't done manual testing myself. Assumming it was tested by the developer)

PR #354, opened 2023-03-01
SLVSCODE-423 Fix displayed message when wrong URL provided during con…

kirill-knize-sonarsource:


src/connected/connectionsetup.ts
```diff
@@ -274,8 +274,8 @@ async function openTokenGenerationPage(message) {
     await connectionSetupPanel.webview.postMessage({ command: 'tokenGenerationPageIsOpen' });
     await vscode.commands.executeCommand(Commands.OPEN_BROWSER, vscode.Uri.parse(accountSecurityUrl));
   } catch(error) {
-    const errorMessage = error.message;
-    await connectionSetupPanel.webview.postMessage({ command: 'tokenGenerationPageIsOpen', errorMessage });
+    await connectionSetupPanel.webview.postMessage({ command: 'tokenGenerationPageIsOpen',
+      errorMessage: 'Incorrect URL or server is not available' });
```
sophio-japharidze-sonarsource, line 278: Maybe `Invalid URL` is more accurate?

jblievremont: `Server at '${cleanedUrl}' could not be reached.` ?

kirill-knize-sonarsource: Agreed to go as is.

PR #358, opened 2023-03-08
SLVSCODE-428 Try stdio transport

jblievremont:


sophio-japharidze-sonarsource (commented):
Is it working? When running locally, I don't see any files being analyzed. Neither Python, JS, nor Java

sophio-japharidze-sonarsource (approved):

PR #363, opened 2023-03-14
SLVSCODE-434 Escape HTML in connection setup view (fix SSF-365)

jblievremont:


src/connected/connectionsetup.ts
```diff
@@ -181,7 +182,7 @@ function renderConnectionSetupPanel(context: vscode.ExtensionContext, webview: v
         <vscode-checkbox id="enableNotifications" ${!initialState.disableNotifications ? 'checked' : ''}>
           Receive notifications from ${serverProductName}
         </vscode-checkbox>
-        <input type="hidden" id="enableNotifications-initial" value="${!initialState.disableNotifications}" />
+        <input type="hidden" id="enableNotifications-initial" value="${!(initialState.disableNotifications)}" />
```
kirill-knize-sonarsource, line 185: Why add parenthesis? Isn't it redundant?

kirill-knize-sonarsource (approved):
LGTM with minor comment.

src/connected/connectionsetup.ts
jblievremont: Yes, and IMHO it makes the priority explicit.

PR #369, opened 2023-03-22
SLVSCODE-443 highlight do and don't in the rule description other con…

sophio-japharidze-sonarsource:


styles/rule.css
```diff
@@ -103,3 +103,19 @@ a {
...
+  content: "\2713";
+  color:green;
+  padding-right:5px;
+}
+
+.other-context-list .dont::before {
+  content: "\2717";
+  color:red;
+  padding-right:5px;
+}
```
jblievremont, line 121: Could we try with CSS variables for the colors, to make them theme-aware?

E.g I noticed that we should be able to use these in the webview:

```css
    --vscode-testing-iconFailed: #f14c4c;
    --vscode-testing-iconErrored: #f14c4c;
    --vscode-testing-iconPassed: #73c991;
```

sophio-japharidze-sonarsource: Great idea, tested and it works.

jblievremont (approved):
LGTM 👍🏻 🚀

PR #371, opened 2023-03-29
SLVSCODE-441 SLVSCODE-442 Utility function to list all files that should be considered for analysis

kirill-knize-sonarsource:


vojtech-suchy-sonarsource (approved):
Looks good 👍 

Kudos for introducing _test/suite/commons.ts_

PR #373, opened 2023-04-11
SLVSCODE-450 make sure folders which are not to be autobind are not s…

vojtech-suchy-sonarsource:
To fix [Sonarlint “Don’t Show Again” doesn’t work for Configure Binding dialog](https://community.sonarsource.com/t/sonarlint-dont-show-again-doesnt-work-for-configure-binding-dialog/88859) issue reported on our community


kirill-knize-sonarsource (approved):
LGTM!

PR #382, opened 2023-04-25
Code highlighting in rule descriptions

sophio-japharidze-sonarsource:
Syntax and diff highlighting


src/rules/code-diff.ts
```diff
@@ -0,0 +1,74 @@
...
+}
+
+export function getExamplesFromDom(document) {
+  const pres = document.querySelectorAll(`pre[data-diff-id]`);
+
+  return (
+    Object.values(
+      groupBy(
+        pres.filter(e => e.getAttribute('data-diff-id') !== undefined),
+        e => e.getAttribute('data-diff-id')
```
vojtech-suchy-sonarsource, line 50: nit: Might be worth extracting the `'data-diff-id'` and `'data-diff-type'` attributes to a constant to highlight they are coming from the analyzers and we don't expect these to change

vojtech-suchy-sonarsource (approved):
LGTM 🚢

styles/rule.css
```diff
@@ -119,3 +124,30 @@ a {
...
   padding-right:5px;
 }
+
+
+.code-difference-scrollable .code-removed {
+  background-color: var(--vscode-diffEditor-removedLineBackground);
+  padding-left: 1.5rem;
+  margin-left: -1.5rem;
+  margin-right: -1.5rem;
+  border-radius: 0px;
```
jblievremont, line 134: Maybe these could be refactored to a separate class since they are shared with `.code-added`.

PR #383, opened 2023-04-26
SLVSCODE-457 Honor IDE exclusion settings when looking for files to scan for hotspots

kirill-knize-sonarsource:


jblievremont (changes requested):
See my comment about the configuration scope - file exclusions can be set at the machine, user, workspace and folder levels.

src/hotspot/hotspots.ts
```diff
@@ -230,8 +230,12 @@ function launchScanForHotspots(languageClient: SonarLintExtendedLanguageClient,
 }
 
 export async function getFilesForHotspotsScan(folderUri: vscode.Uri, globPatterns: string[]): Promise<AnalysisFile[]> {
+  const workspaceFolderConfig = vscode.workspace.getConfiguration(null, folderUri);
```
jblievremont, line 233: 👍🏻

jblievremont (approved):
LGTM 👍🏻

PR #386, opened 2023-04-28
SLVSCODE-468 Don't analyse files larger than 500 kilobytes

kirill-knize-sonarsource:


src/util/util.ts
```diff
@@ -24,6 +25,7 @@ export function startedInDebugMode(process: NodeJS.Process): boolean {
 
 export const extension = vscode.extensions.getExtension('SonarSource.sonarlint-vscode');
 export const packageJson = extension.packageJSON;
+export const HOTSPOTS_FULL_SCAN_FILE_SIZE_LIMIT = 500000;
```
jblievremont, line 28:
```suggestion
export const HOTSPOTS_FULL_SCAN_FILE_SIZE_LIMIT_BYTES = 500_000;
```
(nitpick: I believe that the numeric separator helps with understanding the value, and suffixing with the unit helps with understanding the purpose of the constant)

jblievremont (approved):
See my small comment, otherwise LGTM 👍🏻

PR #392, opened 2023-05-05
SLVSCODE-455 Use "standard" rule description rendering for hotspots opened from SonarQube

jblievremont:


jblievremont:
As agreed with @marco-comi-sonarsource, this is a minimal baby step, we can plan for other changes when we work on setting a hotspot's review status from the IDE.

src/hotspot/hotspotContextPanel.ts
```diff
@@ -48,49 +49,76 @@ const categoryShortToLong = {
...
-      <section class="tab">
-      ${hotspot.rule.fixRecommendations}
-      </section>
-    </main>
     </body>
 </html>`;
 }
+
+function hotspotRuleToDescriptionParams(hotspot: RemoteHotspot) {
+  const { key, name, riskDescription, vulnerabilityDescription, fixRecommendations } = hotspot.rule;
```
sophio-japharidze-sonarsource, line 89: Can we get `languageKey` also here somehow? This would be needed for syntax highlighting

sophio-japharidze-sonarsource (approved):
Definitely a good baby step!

PR #393, opened 2023-05-06
SLVSCODE-474 Check if workspace folder bound to compatible server in the very beginning of the hotspots analysis

kirill-knize-sonarsource:


src/lsp/client.ts
```diff
@@ -81,4 +81,8 @@ export class SonarLintExtendedLanguageClient extends LanguageClient {
   getSuggestedBinding(configScopeId:string, connectionId: string): Promise<protocol.GetSuggestedBindingResponse>{
     return this.sendRequest(protocol.GetSuggestedBinding.type, {configScopeId, connectionId});
   }
+
+  checkLocalDetectionSupported(folderUri: string): Promise<protocol.CheckLocalDetectionSupportedResponse> {
```
jblievremont, line 85: I believe this should mention that this checks for the local detection of hotspots.

jblievremont (changes requested):
See my comment on naming, otherwise LGTM!

jblievremont (approved):
LGTM 👍🏻

PR #395, opened 2023-05-09
SLLS-144 use embedded server from new backend

sophio-japharidze-sonarsource:


src/connected/binding.ts
```diff
@@ -107,6 +108,24 @@ export class BindingService {
     return bindingsPerConnectionId;
   }
 
+  async assistBinding(params: AssistBindingParams) {
```
sophio-japharidze-sonarsource, line 111: This will be called when user has attempted to open hotspot in IDE, there is a connection configured but no binding.

src/connected/connectionsetup.ts
```diff
@@ -28,17 +28,27 @@ const TOKEN_RECEIVED_COMMAND = 'tokenReceived';
 const OPEN_TOKEN_GENERATION_PAGE_COMMAND = 'openTokenGenerationPage';
 const SAVE_CONNECTION_COMMAND = 'saveConnection';
 
+export function assistCreatingConnection(context: vscode.ExtensionContext) {
```
sophio-japharidze-sonarsource, line 31: This will be called when user has tried to open hotspot in IDE but there is no connection configured.

src/extension.ts
```diff
@@ -504,11 +510,15 @@ function installCustomRequestHandlers(context: VSCode.ExtensionContext) {
...
   languageClient.onRequest(protocol.GetTokenForServer.type, serverId => getTokenForServer(serverId));
-  languageClient.onNotification(protocol.SubmitTokenNotification.type, token => handleTokenReceivedNotification(token));
   languageClient.onNotification(protocol.PublishHotspotsForFile.type, hotspotsPerFile => {
     hotspotsTreeDataProvider.refresh(hotspotsPerFile);
     updateSonarLintViewContainerBadge();
   });
+  languageClient.onNotification(protocol.AssistCreatingConnection.type, assistCreatingConnection(context));
+  languageClient.onNotification(
+    protocol.AssistBinding.type,
+    async params => await BindingService.instance.assistBinding(params)
```
sophio-japharidze-sonarsource, line 520: SQ is complaining about this line and I don't understand how to fix it. If I simply put `BindingService.instance.assistBinding`, the method is not being called at all.

jblievremont: I marked it as won't fix (probably should have been false positive actually) - there was a discussion with the JS/TS/CSS squad that this rule raised a lot of FP on SQ and SC so they won't enable it by default.

jblievremont (approved):
LGTM (finally took some time to test it) 👍🏻

PR #410, opened 2023-06-08
SLVSCODE-502 Use global state to store install time

henryju:


jblievremont (commented):
Looks OK from a functional point of view as it only adds new attributes. Could you please also update the telemetry sample to reflect these changes?

jblievremont (approved):
LGTM, thanks! 👍🏻

PR #427, opened 2023-07-03
SLVSCODE-520   Better UX for SSL certificates

kirill-knize-sonarsource:


src/extension.ts
```diff
@@ -565,6 +571,7 @@ function installCustomRequestHandlers(context: VSCode.ExtensionContext) {
     protocol.AssistBinding.type,
     async params => await BindingService.instance.assistBinding(params)
   );
+  languageClient.onRequest(protocol.SslCertificateConfirmation.type, cert => askSslCertificateConfirmation(cert))
```
sophio-japharidze-sonarsource, line 574: Why not call directly the `showSslCertificateConfirmationDialog()`?

sophio-japharidze-sonarsource (approved):

PR #432, opened 2023-07-14
Allow reopening local-only issues

vojtech-suchy-sonarsource:
Add command which allows reopening all local-only issue for current file


jblievremont (approved):
LGTM with a minor comment.

PR #442, opened 2023-07-26
BUILD-3478 update dependency of mend and dogfood tasks

sophio-japharidze-sonarsource:
https://sonarsource.atlassian.net/browse/BUILD-3478

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -167,7 +167,7 @@ test_linux_task:
 
 mend_scan_task:
   depends_on:
-    - test_linux
+    - build
   <<: *ONLY_MAIN_BRANCHES
   eks_container:
     <<: *CONTAINER_DEFINITION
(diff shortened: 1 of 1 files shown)


drautureau-sonarsource (approved):

julien-carsique-sonarsource (changes requested):
The fix can be improved.
The issue is running the mend or dogfood task without a previous run of the build task.
- mend and dogfood were missing an explicit dependency on build
- looking at the only_if condition, we need to consider all the cases; here: the cron build
- it appears that Cirrus is not considering transitive dependencies: the build dependency tree "mend -> linux_qa -> build" is not honored if linux_qa is skipped, then mend is not considered as depending on build.

The latter point can be considered as a bug, we are sharing this with Cirrus support. But we don't need to wait for their answer to fix the current build.

ERRATUM: 
The above is not accurate.

julien-carsique-sonarsource (approved):

PR #446, opened 2023-08-03
BUILD-3523 Fix dogfood.json version is not injected

sebastienvermeille:
# BUILD-3523 Fix dogfood.json version is not injected

## Changes
* export the variable so that it is available for envsubst

## How was it tested ?
* locally I could reproduce the issue, then I manually defined the PROJECT version and it worked using `export`.


jblievremont (approved):
LGTM 👍🏻

PR #451, opened 2023-08-21
[MMF-3284] Exclude specific files from automatic analysis

kirill-knize-sonarsource:


package.json
```diff
@@ -256,6 +256,13 @@
           "markdownDescription": "Files whose name match this [glob pattern](https://docs.oracle.com/javase/tutorial/essential/io/fileOps.html#glob) are considered as test files by analyzers. Most rules are *not* evaluated on test files. Example: `{**/test/**,**/*test*,**/*Test*}`",
           "scope": "resource"
         },
+        "sonarlint.analysisExcludesStandalone": {
+          "order": 13,
+          "type": "string",
+          "default": "",
+          "markdownDescription": "Files whose name match this [glob pattern](https://docs.oracle.com/javase/tutorial/essential/io/fileOps.html#glob) will not be processed by analyzers. When SonarLint is connected to SonarQube or SonarCloud this property will be ignored and all files will be analysed. No rules are evaluated on excluded files. Example: `{**/lib/**,**/*generated*}`",
```
sophio-japharidze-sonarsource, line 263: > When SonarLint is connected to SonarQube or SonarCloud this property will be ignored and all files will be analysed.

It is possible to define file exclusions on server right? So maybe we should say 
> When SonarLint is connected to SonarQube or SonarCloud this property will be ignored and server exclusion settings will be applied.

src/util/showMessage.ts
```diff
@@ -81,3 +81,7 @@ export async function showSslCertificateConfirmationDialog(cert: SslCertificateC
     { modal: true }, dontTrust, trust);
   return dialogResponse === trust;
 }
+
+export function showNoActiveFileOpenWarning() {
+  return window.showWarningMessage('At least one file need to be open to use this command');
```
sophio-japharidze-sonarsource, line 86: need**s** to be open

src/issue/issue.ts
```diff
@@ -45,4 +46,18 @@ export class IssueService {
       code2ProtocolConverter(currentlyOpenFileUri));
   }
 
+  analyseOpenFileIgnoringExcludes() {
+    const editor = VSCode.window.activeTextEditor;
```
sophio-japharidze-sonarsource, line 50: this does not check for Jupyter notebooks. We should also check `activeNotebookEditor`

sophio-japharidze-sonarsource (changes requested):
See indivdual comments. Overall looks good and works good except for jupyter documents

sophio-japharidze-sonarsource (approved):

PR #455, opened 2023-08-23
SLVSCODE-537 COBOL analysis is not a beta feature anymore

jblievremont:


vojtech-suchy-sonarsource (commented):
All these changes look good to me.

We can also remove the `ShowNotificationForFirstCobolIssueNotification` from `protocol.ts`

vojtech-suchy-sonarsource (approved):

PR #457, opened 2023-08-23
SLVSCODE-462 Fix layout of context selection tabs on wide monitors

jblievremont:


jblievremont:
This fix looks too simple to me, especially since the previous modification on this line explicitly set the width to 90%.

I tested with a lot of rule descriptions (with/without context, hotspot, new/old taxonomy, single tab etc) but there can be a few cases where it breaks...

PR #459, opened 2023-08-24
SLVSCODE-540 Clean editor decorations after action on hotspot

jblievremont:


sophio-japharidze-sonarsource (approved):
Looks good, tested couple of scenarios 👍

PR #460, opened 2023-08-24
3.21 analyzer updates

sophio-japharidze-sonarsource:


vojtech-suchy-sonarsource (approved):
The changes look good to me, but let's please wait for the ITS to pass to validate we made correct expectations in tests

PR #461, opened 2023-08-24
SLVSCODE-527 Do not save new connection if the status check fails

vojtech-suchy-sonarsource:


sophio-japharidze-sonarsource (approved):
Looks good, tested and works well too! Please have a look at reported code smells and you can also rebase to latest master

PR #467, opened 2023-08-31
SLVSCODE-568 Use updated LS that allows connected mode analysis when server is unreachable

jblievremont:


jblievremont:
Closing this as `master` already uses a newer version.

PR #470, opened 2023-09-05
Upgrade dependencies

sophio-japharidze-sonarsource:
Upgrade dependencies to accommodate Mend


jblievremont (approved):
LGTM 🚢 

I saw that the Istanbul libraries have been updated so I also tested manually that running `npm run test` and `npm run test-cov` still works on a developer box 👍🏻

PR #491, opened 2023-10-25
SLVSCODE-610 open file-level issues

sophio-japharidze-sonarsource:


src/util/issue.ts
```diff
@@ -31,11 +31,16 @@ export async function adaptLocations(flow: Flow) {
...
+  let startPosition = new vscode.Position(0, 0);
+  let endPosition  = new vscode.Position(0, 0);
+  let range = new vscode.Range(startPosition, endPosition);
   // vscode line positions are 0-based
-  const startPosition = new vscode.Position(startLine - 1, startLineOffset);
-  const endPosition = new vscode.Position(endLine - 1, endLineOffset);
-  const range = new vscode.Range(startPosition, endPosition);
-
+  if (startLine !== 0 && endLine !== 0) {
+    // this is a file-level issue
```
jblievremont, line 39: If I understand correctly, this is the opposite: the issue should be considered at the file level if either `startLine` or `endLine` is missing.

Maybe this could be extracted to a function, e.g `getSafeTextRange`.

Also, it could be unit-tested 😇

jblievremont (commented):
Let's have a quick chat about this one.

jblievremont (approved):
LGTM, keeping in mind that we'll have to release the LS and update to the released version 🚀

PR #502, opened 2023-11-09
SLVSCODE-636 analyze T-SQL files in Connected Mode

sophio-japharidze-sonarsource:


src/rules/syntax-highlight.ts
```diff
@@ -27,6 +27,9 @@ export function sonarToHighlightJsLanguageKeyMapping(sonarLanguageKey: string):
       return 'yaml';
     case 'ipynb':
       return 'python';
+    case 'plsql':
+    case 'tsql':
+      return 'sql';
```
sophio-japharidze-sonarsource, line 32: We probably need to enrich `syntax-highlight.test.ts` :)

jblievremont (approved):
🚀

PR #515, opened 2023-12-15
SLVSCODE-657 Fix SSF-533

jblievremont:


src/connected/connectionsetup.ts
```diff
@@ -47,10 +47,24 @@ export function assistCreatingConnection(context: vscode.ExtensionContext) {
...
 }
 
+export function warnAboutUntrustedServer(context: vscode.ExtensionContext) {
+  return async serverUrl => {
+    const yesOption = 'Connect to this SonarQube server';
+    const reply = await vscode.window.showWarningMessage(
+      `Do you trust this SonarQube server?`,
+      { modal: true, detail: `The server '${serverUrl}' is attempting to setup a connection with SonarLint. Letting SonarLint connect to an untrusted SonarQube server is potentially dangerous.
+
+If you don't trust this server, we recommend to cancel this action and manually setup Connected Mode.` }, yesOption);
```
sophio-japharidze-sonarsource, line 61: I think having a secondary action with 'set up manually' option would be useful here? So it would open a form without pre-populated link.

sophio-japharidze-sonarsource: Also, I think grammatically correct phrase would be "We recommend **cancelling**"

sophio-japharidze-sonarsource (changes requested):
LMK what you think about my comments

src/connected/connectionsetup.ts
sophio-japharidze-sonarsource: I have applied the fix for the second comment and created https://sonarsource.atlassian.net/browse/SLVSCODE-663 for the first one

sophio-japharidze-sonarsource (approved):

PR #519, opened 2024-01-03
Update license headers for 2024

sophio-japharidze-sonarsource:


vojtech-suchy-sonarsource (commented):
There are test/build files which have 2023 or 2022 license headers:
- test/suite/scm.test.ts
- files inside build-sonarlint

Do we need to update those?

vojtech-suchy-sonarsource (approved):

PR #532, opened 2024-02-06
SLVSCODE-688 Show message in Rules view about the effects of local configuration

jblievremont:


src/rules/rules.ts
```diff
@@ -128,3 +131,61 @@ export class AllRulesTreeDataProvider implements VSCode.TreeDataProvider<AllRule
 function byName(r1: Rule, r2: Rule) {
   return r1.name.toLowerCase().localeCompare(r2.name.toLowerCase());
 }
+
+export function setRulesViewMessage(allRulesView: VSCode.TreeView<LanguageNode>) {
+  const folderBindingStates = [... BindingService.instance.bindingStatePerFolder().values()];
+  if (allTrue(folderBindingStates)) {
+    allRulesView.message = 'Changes to this view do not have any effect in this workspace since all folders use Connected Mode.';
+  } else if (allFalse(folderBindingStates)) {
+    allRulesView.message = 'Changes to this view apply are restricted to your personal development environment; please use Connected Mode to share a rule set with your team.';
```
sophio-japharidze-sonarsource, line 140: When user goes to this view, i don't think their first intent is to share the config with others. I would suggest to rephrase like
`Changes to this view are restricted to your personal development environment; to share a rule set with your team, please use Connected Mode.`

sophio-japharidze-sonarsource (commented):
Great! ⭐ 
I just left couple of comments about wording

sophio-japharidze-sonarsource (approved):

PR #538, opened 2024-02-19
SLVSCODE-699 limit binding suggestion notifications to 1

sophio-japharidze-sonarsource:


src/connected/autobinding.ts
```diff
@@ -237,10 +234,11 @@ export class AutoBindingService {
         await this.bindingService.saveBinding(
           bindingSuggestion.sonarProjectKey, bindingSuggestion.connectionId, unboundFolder);
         break;
-      case CHOOSE_MANUALLY_ACTION:
+      case CHOOSE_MANUALLY_ACTION: {
```
thahnen, line 237: Why are the curly braces added? I'm not familiar with TypeScript but I would guess that it is better for cases where multiple statements will follow before the break.
This is also the same question for the braces added below.

On the other hand, I could also guess it is to have the variable `targetConnection` in the local scope only to be discarded outside of the case statement.

thahnen (approved):
LGTM! Know the issue from Eclipse where we display way too many binding notifications as well 😶‍🌫️

src/connected/autobinding.ts
sophio-japharidze-sonarsource: https://next.sonarqube.com/sonarqube/project/issues?id=org.sonarsource.sonarlint.vscode%3Asonarlint-vscode&fixedInPullRequest=538

jblievremont (approved):
LGTM 🚀

jblievremont:
> [3 Fixed issues](https://next.sonarqube.com/sonarqube/project/issues?id=org.sonarsource.sonarlint.vscode%3Asonarlint-vscode&fixedInPullRequest=538)

💯 👍🏻 🤩

PR #545, opened 2024-03-07
Fix test failures due to deadlock in backend initialization

jblievremont:


test/suite/extension.test.ts
```diff
@@ -11,26 +11,16 @@ import * as os from 'os';
...
-    this.timeout(60_000);
-    await util.extension.activate();
-  });
-
   setup(async () => {
     await vscode.commands.executeCommand('workbench.action.closeAllEditors');
     await vscode.commands.executeCommand(Commands.SHOW_SONARLINT_OUTPUT);
   });
 
-  test('Extension should be present', () => {
```
sophio-japharidze-sonarsource, line 30: why are we removing this test?

sophio-japharidze-sonarsource (approved):

PR #551, opened 2024-03-21
SLVSCODE-718 Send .sonarlint/*.json files to autobinding facility

jblievremont:
This is a first version for SLVSCODE-718 that sends all JSON files from the `.sonarlint` folder without any special case for .Net solutions (as discussed recently).


sophio-japharidze-sonarsource (approved):
👏

PR #559, opened 2024-04-09
SLVSCODE-720 allow exporting Connected Mode settings

sophio-japharidze-sonarsource:


package.json
```diff
@@ -832,7 +845,7 @@
       }
     }
   },
-  "main": "./dist/extension",
+  "main": "out/src/extension",
```
jblievremont, line 848: I think this change broke the build.

jblievremont (approved):
LGTM 👍🏻 let's send this to ~~dog~~ catfooding 😻

PR #560, opened 2024-04-09
SLVSCODE-738   The "TLS certificate trust" dialog can show erroneous data

kirill-knize-sonarsource:


jblievremont:
I believe we should have a chat with a PM about the best way to fix this. This change fixes the "default" behavior but the modal will still show incorrect data if SSL parameters are provided (e.g. if users configure a custom trust store).

psylum:
A recent firewall change at my organization has just started triggering this dialog. Would it also be possible to indicate either in the dialog or in the verbose logging what the URL that caused the trust issue actually is? with verbose logging enabled, the output from the extension merely says something like "Request failed."

sophio-japharidze-sonarsource:
Fixed by #616

sophio-japharidze-sonarsource:
@psylum your request is tracked [here](https://sonarsource.atlassian.net/browse/SLCORE-930).

PR #571, opened 2024-05-03
BUILD-4894 Add GitHub workflow that checks release credentials

jblievremont:


.github/workflows/credcheck.yml
```diff
@@ -0,0 +1,58 @@
+name: credentials-sanity-check
+
+on:
+  workflow_dispatch:
```
jblievremont, line 4: This is to trigger it manually; ultimately, we could run it on a fixed schedule (e.g. daily) and notify the squad (?) when it fails.

.github/workflows/credcheck.yml
```diff
@@ -0,0 +1,58 @@
...
+      - name: Use Node 18
+        uses: actions/setup-node@v3
+        with:
+          node-version: '18'
+      - name: Get vault secrets
+        id: secrets
+        uses: SonarSource/vault-action-wrapper@dc8fe04652687f7278f3ecd27905967836bab0eb # tag=2.7.4-1
+        with:
+          secrets: |
+            development/artifactory/token/{REPO_OWNER_NAME_DASH}-private-reader access_token | ARTIFACTORY_ACCESS_TOKEN;
```
jblievremont, line 27: We could get the PAT from Vault here with the proper key (and also update the matching post-release action).

.github/workflows/credcheck.yml
```diff
@@ -0,0 +1,58 @@
...
+        working-directory: ./.github/actions/vsce-publish
+        run: |
+          cp ${GITHUB_WORKSPACE}/.cirrus/.npmrc ./.npmrc
+          npm install
+          npx vsce verify-pat -p "${VSCE_TOKEN}" SonarSource
+
+      - name: Check Microsoft marketplace publisher personal access token
+        if: ${{ !cancelled() }}
+        env:
+          OPENVSX_TOKEN: ${{ secrets.OPENVSX_TOKEN }}
```
jblievremont, line 42: This secret (the OpenVSX publication token) could _also_ be migrated to Vault.

jblievremont (commented):
@tomverin would you like to have a look at this workflow?

The idea is to be able to validate release-time credentials without actually attempting a release.

sophio-japharidze-sonarsource (approved):

.github/workflows/credcheck.yml
```diff
@@ -0,0 +1,58 @@
...
+      - name: Check Artifactory reader token
+        if: ${{ !cancelled() }}
+        env:
+          ARTIFACTORY_URL: https://repox.jfrog.io/artifactory
+          ARTIFACTORY_ACCESS_TOKEN: ${{ fromJSON(steps.secrets.outputs.vault).ARTIFACTORY_ACCESS_TOKEN }}
+        run: |
+          curl -s -o /dev/null \
+            -w '%{http_code}\n' \
+            -H "Authorization: Bearer ${ARTIFACTORY_ACCESS_TOKEN}" \
+            "${ARTIFACTORY_URL}/sonarsource-releases/org/sonarsource/sonarlint/vscode/sonarlint-vscode"
```
tomverin, line 58: you're not pointing to any artifacts, what is the purpose of this command?

tomverin (approved):

.github/workflows/credcheck.yml
jblievremont: Mostly, making sure that the provided token has access to the parent folder of all SLVSCODE releases - it should get a `200` and not a `401` / `403`.

PR #579, opened 2024-05-30
SLLS-234 get project names by key

sophio-japharidze-sonarsource:
With my modifications, users should no longer see a similar `<project not found>` message in the SonarLint Connected Mode view.

809ebf07-d406-44e7-9718-0a677ee97aa4

Unfortunately, they will still not be able to find their project in the list when setting up binding. As a workaround, they can always define the binding setting manually under project/.vscode/settings.json, or also use a [shared binding configuration](https://docs.sonarsource.com/sonarlint/vs-code/team-features/connected-mode/#sharing-your-setup).


jblievremont (approved):
LGTM 👍🏻

PR #580, opened 2024-06-03
Prepare 4.6 release

sophio-japharidze-sonarsource:


CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 * Update IaC analyzer 1.27 -> [1.28](https://github.com/SonarSource/sonar-iac/releases/tag/1.28.0.9889) -> [1.29](https://github.com/SonarSource/sonar-iac/releases/tag/1.29.0.10169) -> [1.30](https://github.com/SonarSource/sonar-iac/releases/tag/1.30.0.10357) -> [1.31](https://github.com/SonarSource/sonar-iac/releases/tag/1.31.0.10579), analysis of Spring configuration files, new rules for ARM, bug fixes and FP fixes
 * Update Python analyzer 4.17 -> [4.18](https://sonarsource.atlassian.net/issues/?jql=fixVersion%20=%2015609%20ORDER%20BY%20created%20ASC), 5 new rules, FP fixes
 * Update C# analyzer 9.23.2 -> [9.24](https://github.com/SonarSource/sonar-dotnet/releases/tag/9.24.0.89429) -> [9.25](https://github.com/SonarSource/sonar-dotnet/releases/tag/9.25.0.90414), 8 new rules, improvements
+* Bugfixes & hardening
```
nquinquenel, line 8: Do you usually mention this? I don't see this in changelog of previous versions

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
...
+* Delegate analysis execution to common SonarLint backend
```
nquinquenel, line 9: Just FYI, I did not mention this in SLI, as I think it does not add/change any functional value for our users.

nquinquenel (approved):
LGTM, just a few remarks

CHANGELOG.md
sophio-japharidze-sonarsource: We mentioned SLOOP migration as "backend refactoring" in one of the previous changelogs. I think it makes sense because it is a "change", even though internal.

I think having this transparency does not hurt.

Indeed we usually don't mention hardening part, i will remove it

nquinquenel: SLOOP migration could be impacting the user as it is starting a new Java process, so I think it's ok to mention it. Here I think it's purely technical and transparent to the user. But it's up to you, I think it's not a big deal 😄

PR #583, opened 2024-06-04
BUILD-5219 use Hashicorp Vault to retrieve secrets

sebastienvermeille:
# BUILD-5219 use Hashicorp Vault to retrieve secrets

## Changes
* Retrieve secrets from Hashicorp Vault instead of the one stored in GitHub repository
  That way the new values are used/updated in Vault.
  
  It should resolve issues of [release](https://github.com/SonarSource/sonarlint-vscode/actions/runs/9349453145) 

## How was it tested?
* I triggered credcheck.yml on this same branch https://github.com/SonarSource/sonarlint-vscode/actions/runs/9369027317

### DIFF
--- a/.github/workflows/credcheck.yml
+++ b/.github/workflows/credcheck.yml
@@ -27,12 +27,14 @@ jobs:
         with:
           secrets: |
             development/artifactory/token/{REPO_OWNER_NAME_DASH}-private-reader access_token | ARTIFACTORY_ACCESS_TOKEN;
+            development/kv/data/visualstudio VSCE_TOKEN | VISUALSTUDIO_PAT;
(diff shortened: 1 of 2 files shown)


jblievremont (commented):
Looks good to me, could you please also update the credentials check workflow in `.github/workflows/credcheck.yml`? This way we can manually check that it gets the correct value from Vault.

sebastienvermeille:
> Looks good to me, could you please also update the credentials check workflow in `.github/workflows/credcheck.yml`? This way we can manually check that it gets the correct value from Vault.

Hi @jblievremont I implemented it in this new commit bdeb830176f8f185cd584c98e525116d3becbdc0

jblievremont (approved):
Nice! And I see that you've already run the credentials check workflow, thanks!

LGTM 🚢

PR #588, opened 2024-06-24
SLVSCODE-747 make organizationKey into a dropdown

sophio-japharidze-sonarsource:
Screenshot 2024-06-24 at 11 26 33

### DIFF
--- a/webview-ui/connectionsetup.js
+++ b/webview-ui/connectionsetup.js
@@ -65,6 +65,9 @@ function onChangeOrganizationKey() {
   if (byId('shouldGenerateConnectionId').value === 'true') {
     byId('connectionId').value = sanitize(byId('organizationKey').value);
   }
+  if (byId('organizationKey').value) {
+    byId('organizationKey').setAttribute('selected', true);
+  }
   toggleSaveConnectionButton();
 }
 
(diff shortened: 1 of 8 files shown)


webview-ui/connectionsetup.js
```diff
@@ -279,6 +290,36 @@ function populateTokenField(token) {
   saveState();
 }
 
+function replaceOrganizationDropdown(organizations) {
+  const dropdown = byId('organizationKey');
+  dropdown.innerHTML = '';
+  if (organizations.length === 0) {
+    // do something to indicate that there are no organizations
```
jblievremont, line 297: 😅 
In this case, one possible thing would be to fall back to a text input. Might be considered "nice to have" since the organization can be set manually in the `settings.json` after creation.

jblievremont (approved):
Manually tested!

See my comment on the "no organization" state: we may want to do something e.g. for people who are trying to use connected mode on a public project from an organization they don't belong to.

webview-ui/connectionsetup.js
sophio-japharidze-sonarsource: I agree, that would be great. I created a followup [hardening ticket](https://sonarsource.atlassian.net/browse/SLVSCODE-785) for this scenario, as it can be tackled independently and is not a blocker for now.

PR #601, opened 2024-07-24
SLVSCODE-809 show fix suggestions

sophio-japharidze-sonarsource:


telemetry-sample.md
```diff
@@ -75,6 +75,19 @@
     "imported_bindings_count": 2,
     "auto_bindings_count": 1,
     "exported_connected_mode_count": 4
+  },
+  "fix_suggestion_received": {
```
nquinquenel, line 79: This will be modified following the feedback from the data team

src/fixSuggestions/fixSuggestionsService.ts
```diff
@@ -0,0 +1,45 @@
...
+		return FixSuggestionService._instance;
+	}
+
+	showFixSuggestion = async (params : ShowFixSuggestionParams) => {
+		try {
+			const fileUri = vscode.Uri.parse(params.fileUri);
+			await vscode.window.showTextDocument(fileUri);
+			const wsedit = new vscode.WorkspaceEdit();
+			params.textEdits.forEach(async (edit) => {
+				const range = new vscode.Range(edit.beforeLineRange.startLine, 0, edit.beforeLineRange.endLine, 10000);
```
jblievremont, line 24: Where does this `10000` come from? (Sh|c)ould it be extracted to a constant with a descriptive name?

telemetry-sample.md
```diff
@@ -75,6 +75,31 @@
...
+	          "count_ai_suggestions": 2,
+            "ai_suggestion_opened_from": "SONARCLOUD",
+            "snippets": [
+                {
+                    "status": "DECLINED",
+                    "snippet_index": null
+                }
+            ]
+        }
+    ]
```
jblievremont, line 103: Nitpick: previous samples use a 2-space indent.

jblievremont (commented):
Looks good on the functional side, please see my 2 small comments on the code.

src/fixSuggestions/fixSuggestionsService.ts
sophio-japharidze-sonarsource: It is just a placeholder so that we don't have to calculate the actual length of the line 🙃
I will put it in a constant

src/fixSuggestions/fixSuggestionsService.ts
```diff
@@ -0,0 +1,46 @@
...
+		return FixSuggestionService._instance;
+	}
+
+	showFixSuggestion = async (params : ShowFixSuggestionParams) => {
+		try {
+			const fileUri = vscode.Uri.parse(params.fileUri);
+			await vscode.window.showTextDocument(fileUri);
+			const wsedit = new vscode.WorkspaceEdit();
+			params.textEdits.forEach(async (edit) => {
+				const range = new vscode.Range(edit.beforeLineRange.startLine, 0, edit.beforeLineRange.endLine, FixSuggestionService.END_OF_LINE_OFFSET);
```
nquinquenel, line 25: What if the start/end lines are wrong? (i.e. outside the bounds of the document)

sophio-japharidze-sonarsource: Good point!! I just pushed a commit to apply change to a "validated range". For example, if we receive a range that is completely out of the file, like line 100 for a 5-line file, the user will see a warning message that "file content has changed and fix might not be applicable anymore", and they will see a fix suggestion at the very end of the file.

nquinquenel (approved):
LGTM

PR #610, opened 2024-08-15
SLVSCODE-827 fix line numbers when opening fix suggestion

sophio-japharidze-sonarsource:


thahnen (approved):
LGTM! Very good catch, have to keep this in mind for Eclipse as well.

PR #611, opened 2024-08-15
SLVSCODE-826 fix automatic connection creation with SonarCloud

sophio-japharidze-sonarsource:


thahnen (approved):
LGTM! This also now looks a bit more streamlined with the strict distinction between SonarClound, SonarQube (with token in >=10.4) and SonarQube (without token in <10.4)!

PR #621, opened 2024-09-16
SLVSCODE-847 Prevent clogging of dev console with messages from directory traversal

jblievremont:


sophio-japharidze-sonarsource (changes requested):
As we discussed, the check should use `&`.

sophio-japharidze-sonarsource (approved):

PR #628, opened 2024-10-01
SLVSCODE-870 Fix silent failure of deployment to OpenVSX

jblievremont:
[SLVSCODE-870](https://sonarsource.atlassian.net/browse/SLVSCODE-870)

[SLVSCODE-870]: https://sonarsource.atlassian.net/browse/SLVSCODE-870?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/ovsx-publish/main.js
+++ b/.github/actions/ovsx-publish/main.js
@@ -21,6 +21,8 @@ const {
   OPENVSX_TOKEN
 } = process.env;
 
+console.log(`Publishing ${ARTIFACT_FILE} to OpenVSX`);
+
 (async () => {
   /**
    * @type ovsx.PublishOptions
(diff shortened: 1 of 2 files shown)


.github/actions/ovsx-publish/main.js
```diff
@@ -29,10 +31,13 @@ const {
     extensionFile: ARTIFACT_FILE,
     pat: OPENVSX_TOKEN
   };
-  await ovsx.publish(options);
-
+  const [ publicationResult ] = await ovsx.publish(options);
```
jblievremont, line 34: This is what caused the silent failure: `ovsx.publish` returns an array of `PromiseSettledResult` instead of failing 🤦🏻

sophio-japharidze-sonarsource: If I'm looking at the [correct thing](https://github.com/eclipse/openvsx/blob/f4b5687f062ada70bfed41cf55ebf36a4b50bb6b/cli/src/publish.ts#L18), `publish()` returns a single `Promise<PromiseSettledResult<void>[]>`. With the `await`, we will get the array of settled/unsettled promises.

So I am thinking that we should consider the full array (not only the first element), and check if any of them have been rejected in the `if` below. 🤔

jblievremont: In theory, yes; and when I run the script, the array contains only a single result. It could contain several items if we were publishing several packages at once.

sophio-japharidze-sonarsource: OK, that makes sense!

sophio-japharidze-sonarsource (approved):
This means all our release actions will start failing "loudly" from now on :D Let's discuss before we merge. Maybe it's best to disable the OVSX publish temporarily until we have a long-term solution

PR #631, opened 2024-10-09
SLVSCODE-871

sophio-japharidze-sonarsource:


package.json
```diff
@@ -873,7 +873,7 @@
     "expand-home-dir": "0.0.3",
     "find-java-home": "1.2.2",
     "follow-redirects": "^1.15.6",
-    "globby": "^13.2.2",
```
sophio-japharidze-sonarsource, line 876: Some of these dependencies are causing Mend alerts. Hoping the update will reduce the number.

jblievremont (approved):
LGTM 🚢

PR #646, opened 2024-11-20
SLVSCODE-891 new icons for standard severity mode

sophio-japharidze-sonarsource:
Screenshot 2024-11-20 at 10 06 43
Screenshot 2024-11-20 at 10 06 35
Screenshot 2024-11-20 at 10 06 22


src/rules/rulepanel.ts
```diff
@@ -111,6 +111,18 @@ function renderImpact(softwareQuality: string, severity: string, resolver: Resou
...
 
+function renderStandardModeSeverityDetails(ruleType: string, severity: string, resolver: ResourceResolver) {
+  const ruleTypeToLowerCase = ruleType.toLocaleLowerCase('en-us');
+  const severityToLowerCase = severity.toLocaleLowerCase('en-us');
+  const ruleTypeImgSrc = resolver.resolve('images', 'standardMode', ruleTypeToLowerCase, `${severityToLowerCase}.svg`);
+  const severityImgSrc = resolver.resolve('images', 'severity', `${severityToLowerCase}.svg`);
+  return `<div class="impact severity-${severityToLowerCase} capsule">
+  <img alt="${clean(ruleType)}" src="${ruleTypeImgSrc}"/>
+  &nbsp;${clean(ruleType)}&nbsp;
+  <img alt="${severityToLowerCase}" src="${severityImgSrc}"/>
```
jblievremont, line 122: Could we imagine using the contents of the SVG files as elements that could be styled with CSS instead of having 15 icons to account for all severities?

jblievremont (commented):
Looks good for a first baby step 🙂 

It would be even better if we could de-duplicate the type icons; I'm not comfortable with having 15 icons only because of different colors.

jblievremont (approved):
Nice 🤩 and works very well with theme switching 💯

PR #659, opened 2024-12-06
SLVSCODE-953 untar eslint-bridge bundle and send its path to slls

sophio-japharidze-sonarsource:


src/extension.ts
```diff
@@ -200,6 +200,7 @@ export async function activate(context: VSCode.ExtensionContext) {
           omnisharpDirectory: Path.resolve(context.extensionPath, 'omnisharp'),
           csharpOssPath: Path.resolve(context.extensionPath, 'analyzers', 'sonarcsharp.jar'),
           csharpEnterprisePath: Path.resolve(context.extensionPath, 'analyzers', 'csharpenterprise.jar'),
+          eslintBridgeServerPath: Path.resolve(context.extensionPath, 'eslint-bridge')
```
jblievremont, line 203: See my comment on SLLS: the `additionalAttributes` field is sent as-is to the telemetry, so I believe it needs to be renamed and cleaned-up.

jblievremont (changes requested):
Looks good on the functional side (and tested manually); let's make sure that we are not sending anything unnecessary in the telemetry payload 😅

jblievremont (approved):
LGTM ⏩!

sophio-japharidze-sonarsource:
Superseded by #661

PR #675, opened 2025-01-03
SLVSCODE-972 try fix flaky autobinding test

sophio-japharidze-sonarsource:


test/suite/autobinding.test.ts
```diff
@@ -439,7 +438,7 @@ suite('Auto Binding Test Suite', () => {
   });
 });
 
-async function deleteSettingsFiles() {
```
sophio-japharidze-sonarsource, line 442: the temporary settings files need to be deleted not to pollute the state for upcoming test suites. At the same time, binding needs to be cleaned up after each test.

Possibly there used to be a race condition between the two actions. This change aims to mitigate this.

PR #681, opened 2025-02-05
SLVSCODE-1004 multi region support

sophio-japharidze-sonarsource:
[SLVSCODE-1004](https://sonarsource.atlassian.net/browse/SLVSCODE-1004)
SLVSCODE-1005
SLVSCODE-1006
SLVSCODE-1008

[SLVSCODE-1004]: https://sonarsource.atlassian.net/browse/SLVSCODE-1004?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/util/util.ts
```diff
@@ -319,3 +320,23 @@ export function getSeverity(severity: number): vscode.DiagnosticSeverity {
       return vscode.DiagnosticSeverity.Warning;
   }
 }
+
+export function sonarCloudRegionToLabel(region: number): SonarCloudRegion {
```
sophio-japharidze-sonarsource, line 324: I decided to implement the conversion here instead of the language server because I would have to rewrite several classes to achieve it there

src/connected/sharedConnectedModeSettingsService.ts
```diff
@@ -110,18 +117,19 @@ export class SharedConnectedModeSettingsService implements FileSystemSubscriber
        configuration files are available to bind folder '${workspaceFolder.name}'
         to a Sonar server. Do you want to use the shared configuration?`;
     const useConfigurationHandler = async () => {
-      const quickPickItems: vscode.QuickPickItem[] = uniqueSuggestions.map(s => {
+      const quickPickItems: CustomQuickPickItem[] = uniqueSuggestions.map(s => {
         return {
           label: s.projectKey,
           description: s.organization || s.serverUrl,
-          detail: s.organization ? 'SonarQube Cloud' : 'SonarQube Server'
+          detail: s.organization ? `[${s.region}] SonarQube Cloud` : 'SonarQube Server',
```
jblievremont, line 124: For consistency, is it OK to always have the region displayed here vs what is done for the tree view items?

src/lsp/protocol.ts
```diff
@@ -601,6 +607,7 @@ export interface AssistCreatingConnectionParams {
   isSonarCloud: boolean;
   serverUrlOrOrganisationKey: string;
   token: string;
+  region?: number;
```
jblievremont, line 610: Is this type expected here? Should it be the same as the one used in `ConnectionSuggestion`?

src/util/dogfoodingUtils.ts
```diff
@@ -0,0 +1,13 @@
...
+ * SonarLint for VisualStudio Code
+ * Copyright (C) 2017-2025 SonarSource SA
+ * sonarlint@sonarsource.com
+ * Licensed under the LGPLv3 License. See LICENSE.txt in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+'use strict';
+
+const DOGFOODING_ENVIRONMENT_VARIABLE_NAME = 'SONARSOURCE_DOGFOODING';
+
+export function isDogfoodingEnvironment() : boolean {
```
jblievremont, line 11: This looks a lot like a duplicate from `src/monitoring/dogfooding.ts` 😇

package.json
```diff
@@ -155,6 +155,10 @@
                 "type": "string",
                 "description": "A unique identifier for this connection to be used as a reference in `#sonarlint.connectedMode.project#`. Only needed if you plan to use multiple connections to SonarQube (Server, Cloud)."
               },
+              "region": {
+                "type": "string",
```
jblievremont, line 159:
```suggestion
                "type": "string",
                "enum": ["EU", "US"],
```
This should constrain the possible values and lower the risk of unacceptable ones.

webview-ui/connectionsetup.js
```diff
@@ -113,20 +113,31 @@ function onClickGenerateToken() {
...
   const serverUrlElement = byId('serverUrl');
-  const serverUrl = serverUrlElement ? serverUrlElement.value : 'https://sonarcloud.io';
+  let serverUrl;
+  let region = null;
+  if (serverUrlElement) {
+    serverUrl = serverUrlElement.value;
+  } else {
+    const regionField = byId('region');
+    region = regionField ? regionField.value : 'EU';
+    serverUrl = region === 'US' ? 'https://us-sc-staging.io/' : 'https://sonarcloud.io';
```
jblievremont, line 123: Is it OK to have `us-sc-staging.io` here?

jblievremont (changes requested):
Looks good from a functional point of view 👍🏻 please check my comments about the implementation 👀

src/lsp/protocol.ts
sophio-japharidze-sonarsource: In fact, the type in ConnectionSuggestion should be a `number` too. I decided to do the conversion completely on the client side.

src/connected/sharedConnectedModeSettingsService.ts
sophio-japharidze-sonarsource: I added the check for Dogfooding Environment. Otherwise, in this case user needs to make an explicit choice, so the more details we display the better (I think). Also, users ending up in this branch is quite a rare case (I think) :D .

jblievremont (approved):
LGTM 👍🏻 🚀 ☁️

PR #692, opened 2025-02-25
SLVSCODE-1039 Clean unused analyzers

jblievremont:


src/cfamily/ondemand.ts
```diff
@@ -7,23 +7,32 @@
...
+}
+
 export function maybeAddCFamilyJar(params: string[]) {
   const expectedVersion: string = util.packageJson.jarDependencies.filter(dep => dep.artifactId === CFAMILY_PLUGIN_ID)[0].version;
-  const onDemandAnalyzersPath = path.resolve(util.extensionPath, '..', 'sonarsource.sonarlint_ondemand-analyzers');
-  const maybeCFamilyJar = path.resolve(onDemandAnalyzersPath, CFAMILY_PLUGIN_ID, expectedVersion, CFAMILY_JAR);
+  const maybeCFamilyJar = path.resolve(getOnDemandAnalyzersPath(), CFAMILY_PLUGIN_ID, expectedVersion, CFAMILY_JAR);
   if (fs.existsSync(maybeCFamilyJar)) {
     params.push(maybeCFamilyJar);
+    util.extensionContext.globalState.update(lastUsedKey(CFAMILY_PLUGIN_ID, expectedVersion), DateTime.now().toMillis());
```
sophio-japharidze-sonarsource, line 31: I think this method returns Thenable? So maybe, in theory, it is possible to quickly remove the currently needed analyzer before the state is updated.

I would suggest to either await this update, or only iterate through the analyzers that are NOT of `expectedVersion` later on.

sophio-japharidze-sonarsource (approved):
Hopefully changing to `async` does not break anything 😅

PR #706, opened 2025-04-01
SLVSCODE-1109 base changes for AI CodeFix

sophio-japharidze-sonarsource:
[SLVSCODE-1109](https://sonarsource.atlassian.net/browse/SLVSCODE-1109)

[SLVSCODE-1109]: https://sonarsource.atlassian.net/browse/SLVSCODE-1109?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -97,6 +97,7 @@ let allHotspotsView: VSCode.TreeView<HotspotTreeViewItem>;
 let helpAndFeedbackTreeDataProvider: HelpAndFeedbackTreeDataProvider;
 let helpAndFeedbackView: VSCode.TreeView<HelpAndFeedbackLink>;
 let taintVulnerabilityCollection: VSCode.DiagnosticCollection;
+let currentProgress: { progress: VSCode.Progress<{ increment?: number }>, resolve: () => void } | undefined;
 
 async function runJavaServer(context: VSCode.ExtensionContext): Promise<StreamInfo> {
   try {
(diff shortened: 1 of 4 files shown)


src/extension.ts
```diff
@@ -97,6 +97,7 @@ let allHotspotsView: VSCode.TreeView<HotspotTreeViewItem>;
 let helpAndFeedbackTreeDataProvider: HelpAndFeedbackTreeDataProvider;
 let helpAndFeedbackView: VSCode.TreeView<HelpAndFeedbackLink>;
 let taintVulnerabilityCollection: VSCode.DiagnosticCollection;
+let currentProgress: { progress: VSCode.Progress<{ increment?: number }>, resolve: () => void } | undefined;
```
jblievremont, line 100: Having a single progress at a time may be dangerous. How about having a map indexed on the `taskId` from the LSP methods?

src/extension.ts
```diff
@@ -270,6 +271,36 @@ export async function activate(context: VSCode.ExtensionContext) {
   IssueService.init(languageClient, secondaryLocationsTree, issueLocationsView);
 
   context.subscriptions.push(languageClient.onNotification(ShowIssueNotification.type, IssueService.showIssue));
+  
+  context.subscriptions.push(languageClient.onNotification(protocol.StartProgressNotification.type, (params: protocol.StartProgressNotificationParams) => {
+    if (currentProgress) {
+      // If there's an existing progress, resolve it first
+      currentProgress.resolve();
```
jblievremont, line 278: Continuing on my comment above: automatically `resolve()`-ing the current progress may not be what is expected: if the end notification was not received for a given `taskId`, the task is probably still going on.

jblievremont (changes requested):
Looks good as a first functional implementation, but I'm a bit worried about the implementation of the progress notification handlers: even if right now this feature is the only place where it is used, it is not guaranteed to be the case in the future. Since the LS is already sending a `taskId` for a given progress job, the client should take it into account.

src/extension.ts
sophio-japharidze-sonarsource: Yes I was planning to tackle this. Did not go forward since I was not sure if LS part was OK. I will fix it.

jblievremont (approved):

PR #726, opened 2025-05-09
SLVSCODE-1164 allow storing token for two connections with the same org key

sophio-japharidze-sonarsource:
[SLVSCODE-1164](https://sonarsource.atlassian.net/browse/SLVSCODE-1164)

[SLVSCODE-1164]: https://sonarsource.atlassian.net/browse/SLVSCODE-1164?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -50,7 +50,7 @@ import { showRuleDescription } from './rules/rulepanel';
 import { AllRulesTreeDataProvider, LanguageNode, RuleNode, setRulesViewMessage, toggleRule } from './rules/rules';
 import { initScm, isIgnoredByScm } from './scm/scm';
 import { isFirstSecretDetected, showNotificationForFirstSecretsIssue } from './secrets/secrets';
-import { ConnectionSettingsService, migrateConnectedModeSettings } from './settings/connectionsettings';
+import { ConnectionSettingsService, migrateConnectedModeSettings, migrateSonarQubeCloudTokens } from './settings/connectionsettings';
 import {
   enableVerboseLogs,
   getCurrentConfiguration,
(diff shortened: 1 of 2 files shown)


src/settings/connectionsettings.ts
```diff
@@ -372,7 +372,8 @@ export function isSonarQubeConnection(connection: BaseConnection): connection is
 }
 
 function getTokenStorageKey(connection: SonarQubeConnection | SonarCloudConnection) {
-  return isSonarQubeConnection(connection) ? connection.serverUrl : connection.organizationKey;
+  const regionPrefix = !isSonarQubeConnection(connection) && connection.region ? `${connection.region}_` : '';
```
sophio-japharidze-sonarsource, line 375: this is for legacy connections who might not have region specifically set.

jblievremont: This will not work for the (arguably very few) people who already have a connection with a region: the existing token will not be found.

Note that it may be acceptable to have these users re-configure their token in this case.

src/settings/connectionsettings.ts
```diff
@@ -372,11 +372,25 @@ export function isSonarQubeConnection(connection: BaseConnection): connection is
...
+export async function migrateSonarQubeCloudTokens() {
+  for (const sqcConnection of ConnectionSettingsService.instance.getSonarCloudConnections()) {
+   const oldTokenStorageKey = sqcConnection.organizationKey;
+   const existingToken = await ConnectionSettingsService.instance.getServerToken(oldTokenStorageKey);
+   if (existingToken) {
+   // store existing token with the new key (prefixed with region)
+   ConnectionSettingsService.instance.storeServerToken(getTokenStorageKey(sqcConnection), existingToken);
+   // delete old token record
+   ConnectionSettingsService.instance.deleteTokenForServer(oldTokenStorageKey);
+   }
```
jblievremont, line 394: Nitpick: indentation is off.

src/extension.ts
```diff
@@ -232,6 +232,8 @@ export async function activate(context: VSCode.ExtensionContext) {
   context.subscriptions.push(taintVulnerabilityCollection);
 
   ConnectionSettingsService.init(context, languageClient);
+  // SLVSCODE-1164; one time migration of SonarQube Cloud tokens to get region prefix
```
jblievremont, line 235:
```suggestion
  // SLVSCODE-1164; one time migration of SonarQube Cloud tokens to set region prefix
```

jblievremont (approved):
Two small details, otherwise LGTM 👍🏻

PR #733, opened 2025-05-23
SLVSCODE-1166 Remove redundant 'ask me later' options

sophio-japharidze-sonarsource:
[SLVSCODE-1166](https://sonarsource.atlassian.net/browse/SLVSCODE-1166)

[SLVSCODE-1166]: https://sonarsource.atlassian.net/browse/SLVSCODE-1166?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/cfamily/cfamily.ts
+++ b/src/cfamily/cfamily.ts
@@ -67,12 +67,11 @@ export function notifyMissingCompileCommands(context: vscode.ExtensionContext) {
     if ((await doNotAskAboutCompileCommandsFlag(context)) || remindMeLaterAboutCompileCommandsFlag) {
       return;
     }
-    const remindMeLaterAction = 'Ask me later';
     const configureCompileCommandsAction = 'Configure compile commands';
     const message = `SonarQube for VS Code is unable to analyze C and C++ file(s) because there is no configured compilation 
       database.`;
     vscode.window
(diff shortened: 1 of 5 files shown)


src/cfamily/cfamily.ts
```diff
@@ -81,9 +79,6 @@ export function notifyMissingCompileCommands(context: vscode.ExtensionContext) {
           case configureCompileCommandsAction:
             configureCompilationDatabase();
             break;
-          case remindMeLaterAction:
```
sophio-japharidze-sonarsource, line 84: From API standpoint, we don't have a way to know the difference between user explicitly ❌ dismissing the notification or just ignoring it. So I could either add the default case and set the flag to `true` in case the notification was dismissed, or remove the check altogether.

In case 1, as soon as user ignores or dismisses the first time they see the notification, they won't see it again. So I think approach 2 (current change) is more optimal.

damien-urruty-sonarsource: The state was probably here for a reason, maybe this notification is triggered for every analysis? I fear if we remove this check, we might get a noisy notification displayed at every analysis. But maybe it's not the case and the check was unnecessary. Could you check where this notification is coming from? The fix could also be to maintain some state on the CORE side instead of here

damien-urruty-sonarsource (commented):
Left a comment about one case, the other cases look good

PR #745, opened 2025-06-12
Merge tags and keywords in extension manifest

sophio-japharidze-sonarsource:
Part of SLVSCODE-1102

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "Notebooks",
     "AI"
   ],
-  "tags": [
+  "keywords": [
     "AI",
     "Education",
     "Vulnerability",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -24,17 +24,15 @@
     "Notebooks",
     "AI"
   ],
-  "tags": [
+  "keywords": [
```
sophio-japharidze-sonarsource, line 27: Screenshot 2025-06-12 at 13 53 17
I think the [docs](https://code.visualstudio.com/api/references/extension-manifest#fields) were updated and the `tags` field is not documented anymore. The tags and keywords were supposed to be merged together on the marketplace page but it appears to not be the case (see screenshot).

Having only keywords should make list of our tags very clear & follow the latest documentation

jblievremont (approved):

PR #760, opened 2025-07-07
SLVSCODE-1175 remove multi-region token migration code

sophio-japharidze-sonarsource:
[SLVSCODE-1175](https://sonarsource.atlassian.net/browse/SLVSCODE-1175)

[SLVSCODE-1175]: https://sonarsource.atlassian.net/browse/SLVSCODE-1175?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gabriela-trutan-sonarsource (approved):
I can see that the migration of the token happened in 4.23. For users who updated to 4.23 the migration will take place. But I think for a user that manually updated from 4.22 to 4.27 (skipping 4.23), the migration won't be applied, so they will have to manually re-create the connection. As this is a very edge case, it should probably be fine, but maybe we can consider documenting this behavior.   

Otherwise, LGTM

PR #764, opened 2025-07-10
SLVSCODE-1276 rename hotspots tool

sophio-japharidze-sonarsource:


package.json
```diff
@@ -936,11 +936,11 @@
     "languageModelTools": [
       {
         "name": "sonarqube_list_potential_security_issues",
-        "displayName": "SonarQube: Security Hotspots",
-        "toolReferenceName": "sonarqube_getSecurityHotspots",
+        "displayName": "SonarQube: List Potential Security Issues",
+        "toolReferenceName": "sonarqube_getPotentialSecurityIssues",
```
nquinquenel, line 940: Tool names are usually written fully in snake case

nquinquenel (approved):

package.json
sophio-japharidze-sonarsource: Screenshot 2025-07-10 at 16 29 23

Just trying to be consistent with the other tools.

PR #767, opened 2025-07-14
SLVSCODE-1262 support displaying notebook issues

sophio-japharidze-sonarsource:
[SLVSCODE-1262](https://sonarsource.atlassian.net/browse/SLVSCODE-1262)

[SLVSCODE-1262]: https://sonarsource.atlassian.net/browse/SLVSCODE-1262?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/findings/findingsTreeDataProvider.ts
+++ b/src/findings/findingsTreeDataProvider.ts
@@ -33,23 +33,23 @@ import {
 
 export class FindingsFileNode extends vscode.TreeItem {
   constructor(
-    public readonly fileOrNotebookUri: string,
+    public readonly fileUri: string,
     public readonly findingsCount: number,
     public readonly category?: 'new' | 'older',
     public readonly isNotebook = false,
     public readonly notebookCellUris?: string[]
   ) {
-    super(getFileNameFromFullPath(fileOrNotebookUri), vscode.TreeItemCollapsibleState.Expanded);
+    super(getFileNameFromFullPath(fileUri), vscode.TreeItemCollapsibleState.Expanded);
     
     this.contextValue = 'findingsFileGroup';
     const categorySuffix = category ? `_${category}` : '';
-    this.id = `${fileOrNotebookUri}${categorySuffix}`;
-    this.resourceUri = vscode.Uri.parse(fileOrNotebookUri);
+    this.id = `${fileUri}${categorySuffix}`;
+    this.resourceUri = vscode.Uri.parse(fileUri);
     
     const specifyWorkspaceFolderName = vscode.workspace.workspaceFolders?.length > 1;
     // no need to compute relative path if file is outside any workspace folder
     this.description = vscode.workspace.workspaceFolders ? getRelativePathFromFullPath(
-      fileOrNotebookUri,
+      fileUri,
       vscode.workspace.getWorkspaceFolder(this.resourceUri),
       specifyWorkspaceFolderName
     ) : '';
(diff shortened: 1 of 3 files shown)


src/findings/findingsTreeDataProvider.ts
```diff
@@ -431,8 +458,17 @@ export class FindingsTreeDataProvider implements vscode.TreeDataProvider<Finding
     return false;
   }
 
-  private getFindingsForFile(fileUri: string, category?: 'new' | 'older'): FindingNode[] {
-    const findings = this.findingsCache.get(fileUri) || [];
+  private getFindingsForFile(fileUriOrNotebookUri: string, category?: 'new' | 'older', notebookCellUris?: string[]): FindingNode[] {
+    let findings: FindingNode[] = [];
+
+    if (notebookCellUris) {
```
vnaskos-sonar: Not sure if it's possible, but would it be correct if a `fileUriOrNotebookUri` is a Notebook but does not have cells? 🤔

src/findings/findingsTreeDataProvider.ts
```diff
@@ -431,8 +458,17 @@ export class FindingsTreeDataProvider implements vscode.TreeDataProvider<Finding
...
+      notebookCellUris.forEach(uri => {
+        findings = [...findings, ...(this.findingsCache.get(uri) || [])];
```
vnaskos-sonar: This can be changed to use flatMap for efficiency

```
findings = notebookCellUris.flatMap(uri => this.findingsCache.get(uri) || []);
```

src/findings/findingsTreeDataProviderUtil.ts
```diff
@@ -134,7 +137,11 @@ export function getContextValueForFinding(source: FindingSource, isAiCodeFixable
     case FindingSource.Latest_SonarQube:
       return isAiCodeFixable ? 'AICodeFixableTaintItem' : 'taintVulnerabilityItem';
     case FindingSource.SonarQube:
-      return isAiCodeFixable ? 'AICodeFixableIssueItem' : 'issueItem';
+      if (isNotebookFinding) {
+        return 'notebookIssueItem';
+      } else {
+        return isAiCodeFixable ? 'AICodeFixableIssueItem' : 'issueItem';
+      }
```
vnaskos-sonar: For readability reasons, I would suggest to change this to

```
      if (isNotebookFinding) {
        return 'notebookIssueItem';
      } else if (isAiCodeFixable) {
        return 'AICodeFixableIssueItem';
      } else { 
        return 'issueItem';
      }
```

src/findings/findingsTreeDataProvider.ts
```diff
@@ -27,27 +27,29 @@ import {
...
   getFilterContextValue,
-  selectAndApplyCodeAction
+  selectAndApplyCodeAction,
+  NOTEBOOK_CELL_URI_SCHEME
 } from './findingsTreeDataProviderUtil';
 
 export class FindingsFileNode extends vscode.TreeItem {
   constructor(
-    public readonly fileUri: string,
+    public readonly fileOrNotebookUri: string,
```
vnaskos-sonar: I think it's fine to revert that back to `fileUri` after all the `notebookUri` is also a file URI. The main reason behind this suggestion is that, I noticed quite a lot of places where we still use the term fileUri (eg. on function parameters) and we pass the fileOrNotebookUri.

src/findings/findingsTreeDataProvider.ts
```diff
@@ -320,7 +326,11 @@ export class FindingsTreeDataProvider implements vscode.TreeDataProvider<Finding
     }
 
     if (element instanceof FindingsFileNode) {
-      return this.getFindingsForFile(element.fileUri, element.category);
+      if (element.isNotebook) {
+        return this.getFindingsForFile(element.fileOrNotebookUri, element.category, element.notebookCellUris);
```
vnaskos-sonar: Here we have an `if` that checks for notebooks, and in the `getFindingsForFile` there is also a similar `if`.

Suggestion:
*  The bottom few lines of the `getFindingsForFile` are related to filtering, so they could be extracted to a separate method `filterFindings(...)`
* Split the `getFindingsForFile` to two separate methods `getFindingsForFile` and `getFindingsForNotebook`
* Finally, in replace this to call the `getFindingsForNotebook`

That way we only check once if the file is a Notebook.

vnaskos-sonar (approved):
Looks good 👌 
Just left some minor nitpicks

PR #768, opened 2025-07-15
MMF-4620 findings view coverage

sophio-japharidze-sonarsource:


src/findings/findingsTreeDataProviderUtil.ts
```diff
@@ -221,3 +220,7 @@ export async function selectAndApplyCodeAction(codeActions: vscode.CodeAction[])
     }
   }
 }
+
+export function isNotebookCellUri(uri: string): boolean {
```
vnaskos-sonar, line 226: I noticed this function is only used in `findingsTreeDataProvider.ts`, it could be in the `FindingsTreeDataProvider` as a private method

src/findings/findingsTreeDataProvider.ts
```diff
@@ -458,7 +459,7 @@ export class FindingsTreeDataProvider implements vscode.TreeDataProvider<Finding
     return false;
   }
 
-  private getFindingsForFile(fileUriOrNotebookUri: string, category?: 'new' | 'older', notebookCellUris?: string[]): FindingNode[] {
+  getFindingsForFile(fileUriOrNotebookUri: string, category?: 'new' | 'older', notebookCellUris?: string[]): FindingNode[] {
```
vnaskos-sonar: Although this is totally fine and works as expected, it is generally not recommended to change the visibility of the method just for unit testing.

The `getFindingsForFile` should be considered part of the public method that uses it, and the tests should be performed on the public method (In this case the tests should be for the `getChildren`).

src/findings/findingsTreeDataProviderUtil.ts
sophio-japharidze-sonarsource: Everything in this file is only used by `findingsTreeDataProvider.ts`. I just wanted to make the file smaller somehow :melt:

PR #794, opened 2025-08-28
SLVSCODE-1338 allow users to disable automatic analysis

sophio-japharidze-sonarsource:


src/extension.ts
```diff
@@ -363,8 +371,11 @@ export async function activate(context: VSCode.ExtensionContext) {
   findingsView = VSCode.window.createTreeView('SonarQube.Findings', {
     treeDataProvider: findingsTreeDataProvider
   });
-
   context.subscriptions.push(findingsView);
+
+  AutomaticAnalysisService.init(automaticAnalysisStatusItem, findingsView);
+  AutomaticAnalysisService.instance.updateAutomaticAnalysisStatusBarAndFindingsViewMessage();
```
damien-urruty-sonarsource: Couldn't we make it work without this singleton? It makes things harder to follow than having a local variable and pass it around

src/extension.ts
```diff
@@ -516,6 +527,17 @@ function registerCommands(context: VSCode.ExtensionContext) {
     })
   );
 
+  context.subscriptions.push(
+    VSCode.commands.registerCommand('SonarLint.AutomaticAnalysis.Enable', () => {
+      VSCode.workspace.getConfiguration('sonarlint')
+              .update('automaticAnalysis', true, VSCode.ConfigurationTarget.Global);
```
damien-urruty-sonarsource, line 534: Don't we need to call `updateAutomaticAnalysisStatusBarAndFindingsViewMessage` here and in the Disable?

.cirrus.yml
```diff
@@ -40,7 +40,7 @@ ec2_instance_definition: &WINVM_DEFINITION
   image: base-windows-jdk17-v*
   platform: windows
   region: eu-central-1
-  type: t3.xlarge
+  type: c6i.xlarge
```
damien-urruty-sonarsource: Why did we need to change this? Is it related to this feature?

damien-urruty-sonarsource (commented):
It's mostly good but I asked for some clarifications before approving

src/extension.ts
sophio-japharidze-sonarsource: This will be handled by config change listener on line 343

.cirrus.yml
sophio-japharidze-sonarsource: I tried increasing the cpu to make the qa pass, but it did not help. I have reverted it.

src/extension.ts
```diff
@@ -104,6 +105,7 @@ let findingsTreeDataProvider: FindingsTreeDataProvider;
 let findingsView: VSCode.TreeView<FindingsTreeViewItem>;
 let helpAndFeedbackTreeDataProvider: HelpAndFeedbackTreeDataProvider;
 let helpAndFeedbackView: VSCode.TreeView<HelpAndFeedbackLink>;
+let automaticAnalysisService: AutomaticAnalysisService;
```
damien-urruty-sonarsource: Couldn't it be a local variable in `activate`?

damien-urruty-sonarsource (approved):
LGTM!

PR #797, opened 2025-09-05
SLVSCODE-1128 ovsx publish platform-specific artifacts

sophio-japharidze-sonarsource:


.github/actions/ovsx-publish/main.js
```diff
@@ -31,6 +36,11 @@ console.log(`Publishing ${ARTIFACT_FILE} to OpenVSX`);
     extensionFile: ARTIFACT_FILE,
     pat: OPENVSX_TOKEN
   };
+
+  // Add target platform for platform-specific publishing (skip for universal)
+  if (TARGET_PLATFORM && TARGET_PLATFORM !== 'universal') {
+    options.target = TARGET_PLATFORM;
```
jblievremont: Type `ovsx.PublishOptions` defines a `targets` field, which is an array of `string` - see [declaration](https://github.com/eclipse/openvsx/blob/7dda317015782aea876bd730f7a9b0e9bffd90a7/cli/src/publish-options.ts#L49).

jblievremont (changes requested):
I believe the current version would fail due to a signature mismatch in the arguments passed to the `ovsx` library.

jblievremont (approved):

PR #799, opened 2025-09-05
SLVSCODE-1344 Support flight recorder

jblievremont:


src/contextManager.ts
```diff
@@ -56,11 +57,16 @@ export class ContextManager {
...
+
   resetAllContexts() {
     vscode.commands.executeCommand('setContext', SOME_CONNECTED_MODE_CONTEXT_KEY, undefined);
     vscode.commands.executeCommand('setContext', SOME_STANDALONE_MODE_CONTEXT_KEY, undefined);
     vscode.commands.executeCommand('setContext', HAS_EXPLORED_ISSUE_LOCATIONS_CONTEXT_KEY, undefined);
     vscode.commands.executeCommand('setContext', SHOULD_SHOW_GET_STARTED_VIEW, undefined);
+    vscode.commands.executeCommand('setContext', FLIGHT_RECORDER_RUNNING, undefined);
   }
 
 }
```
sophio-japharidze-sonarsource: missing newline

src/lsp/protocol.ts
```diff
@@ -881,4 +881,16 @@ export namespace NotifyInvalidToken {
   export const type = new lsp.NotificationType<NotifyInvalidTokenParams>('sonarlint/notifyInvalidToken');
 }
 
+export interface FlightRecorderStartedParams {
+  sessionId: string;
+}
+
+export namespace FlightRecorderStartedNotification {
```
sophio-japharidze-sonarsource: I might be mistaken but these two notifications might need to be in different regions - one is from server to client, other is from client to server. Although to be fair we have not been that diligent in keeping the regions clearly separated :(

test/suite/flightrecorder.test.ts
```diff
@@ -0,0 +1,36 @@
...
+    await sleep(1000);
+
+    // Emulate click on "Copy Session ID" (no API to use the actual button)
+    await vscode.commands.executeCommand(Commands.COPY_FLIGHT_RECORDER_SESSION_ID);
+
+    await sleep(1000);
+
+    expect(await vscode.env.clipboard.readText()).to.equal(sessionId);
+
+    await vscode.commands.executeCommand(Commands.DUMP_BACKEND_THREADS);
```
sophio-japharidze-sonarsource, line 33: Is this call only for coverage or does it make any difference?

sophio-japharidze-sonarsource (approved):
Looks good; I think we should finalise the 'prevent abuse' discussion before merging 😄

test/suite/flightrecorder.test.ts
jblievremont: This call is only for coverage; there may be a way to spy on the language client to check that the notification is sent to the LS, but it would be hard to check that a thread dump is actually captured (and the flight recorder is not _actually_ active in the test anyway).

PR #805, opened 2025-09-22
SLVSCODE-1366 automatic mcp setup

sophio-japharidze-sonarsource:


src/mcpServerConfig.ts
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    const sonarQubeMCPConfig = await languageClient.getMCPServerConfiguration(connection.id, token);
+
+    writeSonarQubeMCPConfig(JSON.parse(sonarQubeMCPConfig.jsonConfiguration));
+
+    openMCPServersListIfCursor();
+
+    const openFile = await vscode.window.showInformationMessage(
+      `SonarQube MCP server configured for "${connection.label}"`,
```
nquinquenel:
```suggestion
      `SonarQube MCP Server configured for "${connection.label}"`,
```

src/mcpServerConfig.ts
```diff
@@ -0,0 +1,190 @@
...
+}
+
+function getCurrentSonarQubeMCPServerConfig(): MCPServerConfig {
+  const currentIDE = getCurrentAgentSupportedIDE();
+  const configPath = getMCPConfigPath();
+  const config = readMCPConfig(configPath);
+  return currentIDE === IDE.VSCODE || currentIDE === IDE.VSCODE_INSIDERS
+    ? (config as MCPConfigurationVSCode).servers.sonarqube
+    : (config as MCPConfigurationOthers).mcpServers.sonarqube;
+}
```
nquinquenel: Shouldn't we handle the case where `currentIde` is undefined?

src/mcpServerConfig.ts
nquinquenel: And same in other location (this is the official name 😄 )

src/settings/connectionsettings.ts
```diff
@@ -270,6 +270,23 @@ export class ConnectionSettingsService {
...
+    try {
+      const isSonarQube = connection.contextValue === 'sonarqubeConnection';
+      const connectionDetail = isSonarQube
+        ? await ConnectionSettingsService.instance.loadSonarQubeConnection(connection.id)
+        : await ConnectionSettingsService.instance.loadSonarCloudConnection(connection.id);
+      if (!connectionDetail) {
+        throw new Error(`Could not find SonarQube Server connection with ID: ${connection.id}`);
+      }
+
+      return connectionDetail.token || '';
```
damien-urruty-sonarsource: I'm not sure about the last bit `|| ''`. Wouldn't it fail anyway somewhere else if the token is not set? Shouldn't we fail harder?

src/mcpServerConfig.ts
```diff
@@ -0,0 +1,193 @@
...
+  WINDSURF = 'windsurf',
+  VSCODE_INSIDERS = 'vscode-insiders'
+}
+
+function isCopilotInstalledAndActive(): boolean {
+  const copilotExtension = vscode.extensions.getExtension('GitHub.copilot');
+  return copilotExtension?.isActive;
+}
+
+export function getCurrentAgentSupportedIDE(): IDE | undefined {
```
damien-urruty-sonarsource: The name was a bit blurry for me. May I suggest something else, like `getCurrentIdeWithMcpSupport`

src/mcpServerConfig.ts
```diff
@@ -0,0 +1,193 @@
...
+}
+
+export async function configureMCPServer(
+  connection: Connection,
+  languageClient: SonarLintExtendedLanguageClient
+): Promise<void> {
+  try {
+    const token = await ConnectionSettingsService.instance.getTokenForConnection(connection);
+
+    if (!token) {
```
damien-urruty-sonarsource, line 140: Related to my previous comment, this might be dead code

src/mcpServerConfig.ts
```diff
@@ -0,0 +1,193 @@
...
+      }
+    }
+
+    const sonarQubeMCPConfig = await languageClient.getMCPServerConfiguration(connection.id, token);
+
+    writeSonarQubeMCPConfig(JSON.parse(sonarQubeMCPConfig.jsonConfiguration));
+
+    openMCPServersListIfCursor();
+
+    const openFile = await vscode.window.showInformationMessage(
```
damien-urruty-sonarsource, line 158: Shouldn't this be made only if it's not cursor?

src/mcpServerConfig.ts
sophio-japharidze-sonarsource: This condition will be true in case `token` is an empty string. It will display the notificaion and allow the user to still set up MCP config but with empty token (so it will not work, but it's user's choice). I don't think we should fail harder than this. WDYT?

src/mcpServerConfig.ts
sophio-japharidze-sonarsource: Good question. Even in Cursor there is a config file that user might want to navigate to, so I don't know ...

sophio-japharidze-sonarsource: Since it's not an automatic action and it's a button that user chooses to click, I would keep it. WDYT?

src/settings/connectionsettings.ts
sophio-japharidze-sonarsource: If token is empty, it will give a choice to the user if they want to still proceed. Of course, MCP server will not work properly, if they choose to proceed.

damien-urruty-sonarsource: OK, I just learned that `if (!token)` will evaluate to false if token is an empty string. I would suggest to slightly adjust `getTokenForConnection` by returning a Promise<string?>, I think it makes it clearer that no token can be found instead of relying on the empty string

src/mcpServerConfig.ts
damien-urruty-sonarsource: Yep, let's try like this

PR #807, opened 2025-09-26
Slvscode 1372 ai agents config view

sophio-japharidze-sonarsource:


package.json
```diff
@@ -411,6 +411,12 @@
         "title": "Configure the compilation database for C and C++ analysis",
         "category": "SonarQube"
       },
+      {
+        "command": "SonarLint.RefreshAIAgentsConfiguration",
```
sophio-japharidze-sonarsource, line 415: It is too complicated to listen to both MCP config file changes & the rules file changes across the different IDEs, etc. So I went for an explicit refresh button.

src/aiAgentsConfiguration/aiAgentRuleConfig.ts
```diff
@@ -0,0 +1,99 @@
...
+ * sonarlint@sonarsource.com
+ * Licensed under the LGPLv3 License. See LICENSE.txt in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+'use strict';
+
+import * as vscode from 'vscode';
+import { SonarLintExtendedLanguageClient } from '../lsp/client';
+
+export async function introduceSonarQubeRulesFile(languageClient: SonarLintExtendedLanguageClient): Promise<void> {
+  const workspaceFolder = vscode.workspace.workspaceFolders?.[0];
```
damien-urruty-sonarsource, line 13: Shouldn't we create the rule file in each workspace folder?

src/aiAgentsConfiguration/aiAgentRuleConfig.ts
```diff
@@ -0,0 +1,99 @@
...
+    try {
+      await vscode.workspace.fs.stat(rulesFileUri);
+      const overwrite = await vscode.window.showWarningMessage(
+        'The sonarqube_rules.mdc file already exists. Do you want to overwrite it?',
+        'Overwrite'
+      );
+      if (overwrite !== 'Overwrite') {
+        return;
+      }
+    } catch {}
```
damien-urruty-sonarsource: Should we at least display a log here?

src/aiAgentsConfiguration/aiAgentRuleConfig.ts
```diff
@@ -0,0 +1,99 @@
...
+ * Copyright (C) 2017-2025 SonarSource SA
+ * sonarlint@sonarsource.com
+ * Licensed under the LGPLv3 License. See LICENSE.txt in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+'use strict';
+
+import * as vscode from 'vscode';
+import { SonarLintExtendedLanguageClient } from '../lsp/client';
+
+export async function introduceSonarQubeRulesFile(languageClient: SonarLintExtendedLanguageClient): Promise<void> {
```
damien-urruty-sonarsource, line 12: This method seems to be specific for Cursor. We could maybe put it in the name?

damien-urruty-sonarsource: It should evolve over time, and there is an if in the caller to check if the running IDE is Cursor, so it can stay like this for now

src/aiAgentsConfiguration/aiAgentRuleConfig.ts
damien-urruty-sonarsource: This might be a potential improvement later

PR #945, opened 2026-02-19
DO NOT MERGE: fix_AZlxXlnPSEbp2GJiCGq-_typescript:S7745

revanshu-paliwal-sonarsource:
**Rule:** `typescript:S7745`
**Severity:** MINOR
**Issue description:** `The empty check is useless as `Array#every()` returns `true` for an empty array.`

Remove redundant length check before Array#every()


eray-felek-sonarsource:
Pass: Correct

PR #946, opened 2026-02-19
DO NOT MERGE: fix_AZlxXlrTSEbp2GJiCGsK_typescript:S7757

revanshu-paliwal-sonarsource:
**Rule:** `typescript:S7757`
**Severity:** MINOR
**Issue description:** `Prefer class field declaration over `this` assignment in constructor for static values.`

Move this.rootItem = null to class field declaration


eray-felek-sonarsource:
Pass: A good solution compared to other similar PR addressing the same issue since it does not do unrelated changes

PR #948, opened 2026-02-19
DO NOT MERGE: fix_AZlxXlioSEbp2GJiCGqI_typescript_S7718

revanshu-paliwal-sonarsource:
## 🔧 SonarQube Issue Fix

**Rule:** `typescript:S7718`
**Severity:** MINOR
**Issue description:** `The catch parameter 'notAuri' should be named 'error_'.`

### 📋 Fix Description

Rename catch parameter notAuri to error_

Made with [Cursor](https://cursor.com)


eray-felek-sonarsource:

PR #953, opened 2026-02-20
NO-JIRA [DO NOT MERGE] fix_AZlxXlioSEbp2GJiCGqI

manuel-vianello-sonarsource:
## SonarQube Issue Fix

**Rule:** typescript:S7718 (MINOR)
**Message:** The catch parameter `notAuri` should be named `error_`.

[View issue in SonarCloud](https://sonarcloud.io/project/issues?id=SonarSource_sonarlint-vscode&issues=AZlxXlioSEbp2GJiCGqI&open=AZlxXlioSEbp2GJiCGqI)


eray-felek-sonarsource:

PR #983, opened 2026-02-25
SLVSCODE-1607 DO_NOT_MERGE_fix_AZlxXlioSEbp2GJiCGqI

manuel-vianello-sonarsource:
<strong>typescript:S7718</strong> - The catch parameter `notAuri` should be named `error_`. • <strong>MINOR</strong> • <a href="https://sonarcloud.io/project/issues?id=SonarSource_sonarlint-vscode&issues=AZlxXlioSEbp2GJiCGqI&open=AZlxXlioSEbp2GJiCGqI">View issue</a>


manuel-vianello-sonarsource:

PR #1013, opened 2026-04-16
Revert "SLVSCODE-1443 exclude omnisharp from ovsx targeted packages"

sophio-japharidze-sonarsource:


kirill-knize-sonarsource (approved):
LGTM

PR #1016, opened 2026-04-16
Reapply "SLVSCODE-1443 exclude omnisharp from ovsx targeted packages"

sophio-japharidze-sonarsource:
This reverts commit e695e28929e3c2b1b60025c0da719d2deadc3326.


sophio-japharidze-sonarsource:
Master is green now. must have been a microsoft glitch. Closing this as there is no need to revert the changes

PR #1155, opened 2026-09-16
SLVSCODE-1920 Handle floating promises to restore the Quality Gate

damien-urruty-sonarsource:
## Summary
- Await, return, or explicitly ignore (`void`) floating promises flagged by the newly activated S9383 rule
- Restore the master Quality Gate after SonarQube 2026.5 dropped overall Reliability to C


src/fileSystem/fileSystemServiceImpl.ts
```diff
@@ -61,6 +61,6 @@ export class FileSystemServiceImpl implements FileSystemService {
   }
 
   async didAddWorkspaceFolder(folder: vscode.WorkspaceFolder) {
-    this.crawlDirectory(folder.uri);
+    await this.crawlDirectory(folder.uri);
```
nquinquenel: Shouldn't it be `void` here? It's blocking right now

src/connected/binding.ts
```diff
@@ -341,7 +339,7 @@ export class BindingService {
       if (remoteProjects.size === 0) {
         VSCode.window.showWarningMessage('No remote projects to display.', BIND_MANUALLY_ACTION).then(async action => {
           if (action === BIND_MANUALLY_ACTION) {
-            bindManuallyAction(workspaceFolder);
+            await bindManuallyAction(workspaceFolder);
           }
         });
```
nquinquenel: I think we can do same as with `proposeSharingConfig` above, and flatten to `const action = awiat showWarningMessage`

src/cfamily/cfamily.ts
```diff
@@ -72,13 +72,13 @@ export function notifyMissingCompileCommands(context: vscode.ExtensionContext) {
       database.`;
     vscode.window
       .showWarningMessage(message, configureCompileCommandsAction, DONT_ASK_AGAIN_ACTION)
-      .then(selection => {
+      .then(async selection => {
         switch (selection) {
           case DONT_ASK_AGAIN_ACTION:
             context.workspaceState.update(DO_NOT_ASK_ABOUT_COMPILE_COMMANDS_FLAG, true);
```
nquinquenel: It looks suspicious, shouldn't this be awaited? or similarly to `proposeSharingConfig`

src/languageModelTools/analyzeFileTool.ts
```diff
@@ -35,7 +35,7 @@ export class AnalyzeFileTool implements vscode.LanguageModelTool<IAnalyzeFilePar
       preview: true
     });
     // Analyze the file
-    IssueService.instance.analyseOpenFileIgnoringExcludes(false, textDocument);
+    await IssueService.instance.analyseOpenFileIgnoringExcludes(false, textDocument);
     // Focus on the SonarQube Findings view
     vscode.commands.executeCommand('SonarQube.Findings.focus');
```
nquinquenel: We await analysis but not the opening of the findings view, is that expected?

src/connected/binding.ts
```diff
@@ -238,7 +238,7 @@ export class BindingService {
       remoteProjectsQuickPick.onDidChangeSelection(selection => {
         selectedRemoteProject = selection[0];
 
-        this.saveManualBinding(selectedRemoteProject.description, workspaceFolder, connectionId)
+        void this.saveManualBinding(selectedRemoteProject.description, workspaceFolder, connectionId)
```
nquinquenel:
```suggestion
        void this.saveManualBinding(selectedRemoteProject.description, workspaceFolder, connectionId);
```

nquinquenel (commented):
I'm a bit skeptic about some of the await and the surrounding flows 🤔

src/fileSystem/fileSystemServiceImpl.ts
damien-urruty-sonarsource: Makes sense, yes

src/connected/binding.ts
damien-urruty-sonarsource: Yes. Dialog is awaited in a helper, caller uses void so the list still returns. Done.

src/cfamily/cfamily.ts
damien-urruty-sonarsource: Flatten. The handler already waits on the user’s choice; the caller stays non-blocking. Done.

src/languageModelTools/analyzeFileTool.ts
damien-urruty-sonarsource: Yes, that’s expected. Analysis matters; focusing the view is UI, now void. Done.

src/connected/binding.ts
damien-urruty-sonarsource: Done

src/fileSystem/fileSystemServiceImpl.ts
```diff
@@ -61,6 +61,6 @@ export class FileSystemServiceImpl implements FileSystemService {
   }
 
   async didAddWorkspaceFolder(folder: vscode.WorkspaceFolder) {
-    this.crawlDirectory(folder.uri);
+    void this.crawlDirectory(folder.uri);
   }
```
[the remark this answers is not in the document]
damien-urruty-sonarsource: Makes sense. `void` in `extension.ts` is what keeps folder-add non-blocking. `await` inside the method would be a clearer contract for future callers. Leaving it as-is for this PR.

src/connected/autobinding.ts
nquinquenel, line 248: `autoBindAllFolders` awaits this, but the `.then` is not returned, so the next folder’s prompt still fires immediately. May be worth flattening like `promptToAutoBindMultiChoice` / `promptToBindManually`

nquinquenel (approved):
LGTM! Just one improvement that could be worth doing

src/connected/autobinding.ts
damien-urruty-sonarsource: Done — `promptToBindManually` now awaits the dialog, so `autoBindAllFolders` waits before prompting the next folder.

src/connected/autobinding.ts
```diff
@@ -228,23 +228,20 @@ export class AutoBindingService implements FileSystemSubscriber {
...
+    );
+    if (action === DONT_ASK_AGAIN_ACTION) {
+      await this.workspaceState.update(DO_NOT_ASK_ABOUT_AUTO_BINDING_FOR_FOLDER_FLAG, [
+        ...this.getFoldersThatShouldNotBeAutoBound(),
+        unboundFolder.uri.toString()
+      ]);
+    } else if (action === BIND_ACTION) {
+      const targetConnection = await this.getTargetConnectionForManualBinding();
+      await this.bindingService.createOrEditBinding(targetConnection.connectionId, targetConnection.contextValue);
+    }
```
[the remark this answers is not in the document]
damien-urruty-sonarsource: Done. Guarded all four prompt paths so a cancelled connection pick just returns instead of throwing.

src/connected/autobinding.ts
```diff
@@ -228,23 +228,20 @@ export class AutoBindingService implements FileSystemSubscriber {
...
-          await this.bindingService.createOrEditBinding(targetConnection.connectionId, targetConnection.contextValue);
-        }
-      });
+    const action = await vscode.window.showInformationMessage(
+      CONFIGURE_BINDING_MANUALLY_PROMPT_MESSAGE,
+      BIND_ACTION,
+      DONT_ASK_AGAIN_ACTION
+    );
+    if (action === DONT_ASK_AGAIN_ACTION) {
+      await this.workspaceState.update(DO_NOT_ASK_ABOUT_AUTO_BINDING_FOR_FOLDER_FLAG, [
```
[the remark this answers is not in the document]
damien-urruty-sonarsource: Done. Flattened `askUserBeforeAutoBinding` the same way and awaited the state update.

